PR #53, opened 2018-06-07
fix for travis build: tests stumbled on time zone issues

jo-pol:
Fixes EASY-

#### When applied it will
* 
* 
* 


rvanheest:
I think this is an easier solution, while still keeping the `shouldBe` instead of this less preferable `should startWith`

https://github.com/dans-knaw/easy-bag-index/blob/7b13c6bb109771ba9974b25f7dcc80dc39ff48ad/src/test/scala/nl/knaw/dans/easy/bagindex/server/BagIndexServletSpec.scala#L48

PR #54, opened 2018-06-08
EASY-1487 Transform JSON object to mandatory part of DDM (dataset.xml)

jo-pol:
Fixes [EASY-1487](https://drivenbydata.atlassian.net/browse/EASY-1487)

#### When applied it will have
* [ ] unit tests for each line of the specs
* [ ] unit tests that validate the produce DDM against the schema
* 

The (for now) dropped relations/subjects may illustrate the cause of some choices https://github.com/DANS-KNAW/easy-deposit-api/commit/14defa050c0c052ba8f4aad2e6108506ec0987e6

not applicable

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/AgreementsXml.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/AgreementsXml.scala
@@ -1,0 +1,31 @@
+package nl.knaw.dans.easy.deposit.docs
+
+import org.joda.time.DateTime
+
+import scala.util.{ Failure, Success, Try }
+import scala.xml.Elem
+
+object AgreementsXml {
+  def apply(userId: String, dateSubmitted: DateTime, dm: DatasetMetadata): Try[Elem] = {
+    for {
+      _ <- dm.licenceAccepted
+      privacy <- dm.privacyBoolean
+    } yield
+      <agr:agreements
+          xmlns:agr="http://easy.dans.knaw.nl/schemas/bag/metadata/agreements/"
+          xmlns:dcterms="http://purl.org/dc/terms/"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://easy.dans.knaw.nl/schemas/bag/metadata/agreements/ agreements.xsd">
+        <agr:licenseAgreement>
+          <agr:depositorId>{userId}</agr:depositorId>
+          <dateAccepted>{dateSubmitted}</dateAccepted>
+          <agr:licenseAgreementAccepted>{dm.acceptLicenseAgreement}</agr:licenseAgreementAccepted>
+        </agr:licenseAgreement>
+        <agr:personalDataStatement>
+          <agr:signerId>{userId}</agr:signerId>
+          <agr:dateSigned>{dateSubmitted}</agr:dateSigned>
+          <agr:containsPrivacySensitiveData>{privacy}</agr:containsPrivacySensitiveData>
+        </agr:personalDataStatement>
+      </agr:agreements>
+  }
+}
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/AgreementsXml.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/AgreementsXml.scala
@@ -1,3 +1,18 @@
+/**
+ * Copyright (C) 2018 DANS - Data Archiving and Networked Services (info@dans.knaw.nl)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package nl.knaw.dans.easy.deposit.docs
 
 import org.joda.time.DateTime
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
@@ -58,6 +58,10 @@ case class DatasetMetadata(identifiers: Option[Seq[SchemedValue[String]]] = None
     case SchemedValue(`doiScheme`, value) => value
   })
 
+  lazy val dateCreated: Option[String] = dates.flatMap(_.collectFirst {
+    case QualifiedSchemedValue(Some(`dateCreatedQualifier`), value, _) => value
+  })
+
   lazy val privacyBoolean: Try[Boolean] = privacySensitiveDataPresent match {
     case PrivacySensitiveDataPresent.yes => Success(true)
     case PrivacySensitiveDataPresent.no => Success(false)
@@ -77,6 +81,7 @@ object DatasetMetadata {
   def apply(input: JsonInput): Try[DatasetMetadata] = input.deserialize[DatasetMetadata]
 
   private val doiScheme = "id-type:DOI"
+  private val dateCreatedQualifier = "id-type:DOI"
 
   def missingValue(label: String): InvalidDocumentException = {
     InvalidDocumentException(s"Please set $label in DatasetMetadata")
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
@@ -58,10 +58,6 @@ case class DatasetMetadata(identifiers: Option[Seq[SchemedValue[String]]] = None
     case SchemedValue(`doiScheme`, value) => value
   })
 
-  lazy val dateCreated: Option[String] = dates.flatMap(_.collectFirst {
-    case QualifiedSchemedValue(Some(`dateCreatedQualifier`), value, _) => value
-  })
-
   lazy val privacyBoolean: Try[Boolean] = privacySensitiveDataPresent match {
     case PrivacySensitiveDataPresent.yes => Success(true)
     case PrivacySensitiveDataPresent.no => Success(false)
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
@@ -21,12 +21,9 @@ import nl.knaw.dans.easy.deposit.docs.DatasetMetadata.PrivacySensitiveDataPresen
 import nl.knaw.dans.easy.deposit.docs.DatasetMetadata.RelationQualifier.RelationQualifier
 import nl.knaw.dans.easy.deposit.docs.DatasetMetadata._
 import nl.knaw.dans.easy.deposit.docs.JsonUtil.{ InvalidDocumentException, RichJsonInput, toJson }
-import org.joda.time.DateTime
-import org.joda.time.format.ISODateTimeFormat
 import org.json4s.JsonInput
 
 import scala.util.{ Failure, Success, Try }
-import scala.xml.Elem
 
 case class DatasetMetadata(identifiers: Option[Seq[SchemedValue[String]]] = None,
                            languageOfDescription: Option[SchemedKeyValue[String]] = None,
@@ -40,7 +37,7 @@ case class DatasetMetadata(identifiers: Option[Seq[SchemedValue[String]]] = None
                            alternativeIdentifiers: Option[Seq[SchemedValue[String]]] = None,
                            relations: Option[Seq[RelationType]] = None,
                            languagesOfFiles: Option[Seq[PossiblySchemedKeyValue[String]]] = None,
-                           dates: Option[Seq[Date]] = None,
+                           dates: Option[Seq[QualifiedSchemedValue[String, DateQualifier]]] = None,
                            sources: Option[Seq[String]] = None,
                            instructionsForReuse: Option[Seq[String]] = None,
                            publishers: Option[Seq[String]] = None,
@@ -61,50 +58,18 @@ case class DatasetMetadata(identifiers: Option[Seq[SchemedValue[String]]] = None
     case SchemedValue(`doiScheme`, value) => value
   })
 
+  lazy val privacyBoolean: Try[Boolean] = privacySensitiveDataPresent match {
+    case PrivacySensitiveDataPresent.yes => Success(true)
+    case PrivacySensitiveDataPresent.no => Success(false)
+    case PrivacySensitiveDataPresent.unspecified => Failure(missingValue("PrivacySensitiveDataPresent"))
+  }
+
+  lazy val licenceAccepted: Try[Unit] = if (acceptLicenseAgreement) Success(())
+                                        else Failure(DatasetMetadata.missingValue("AcceptLicenseAgreement"))
+
   def setDoi(value: String): DatasetMetadata = {
     val ids = identifiers.getOrElse(Seq.empty).filter(_.scheme == doiScheme)
     this.copy(identifiers = Some(ids :+ SchemedValue(doiScheme, value)))
-  }
-
-  private lazy val submitDate: Option[String] = {
-    dates.flatMap(_.find(_.qualifier == DateQualifier.dateSubmitted)).map(_.value)
-  }
-
-  lazy val xml: Try[Elem] = Success(<stub/>) // TODO
-
-  def setDateSubmitted(): Try[DatasetMetadata] = {
-    if (submitDate.isDefined)
-      Failure(new Exception("dateSubmitted should not be present"))
-    else {
-      val submitted = Date(DateTime.now(), DateQualifier.dateSubmitted)
-      val newDates = submitted +: dates.getOrElse(Seq.empty)
-      Success(copy(dates = Some(newDates)))
-    }
-  }
-
-  def agreements(userId: String): Try[Elem] = {
-    for {
-      _ <- if (acceptLicenseAgreement) Success(())
-           else Failure(missingValue("AcceptLicenseAgreement"))
-      privacy <- toBoolean(privacySensitiveDataPresent)
-      date = submitDate.getOrElse(throw new IllegalArgumentException("no submitDate"))
-    } yield
-      <agr:agreements
-          xmlns:agr="http://easy.dans.knaw.nl/schemas/bag/metadata/agreements/"
-          xmlns:dcterms="http://purl.org/dc/terms/"
-          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-          xsi:schemaLocation="http://easy.dans.knaw.nl/schemas/bag/metadata/agreements/ agreements.xsd">
-        <agr:licenseAgreement>
-          <agr:depositorId>{userId}</agr:depositorId>
-          <dateAccepted>{date}</dateAccepted>
-          <agr:licenseAgreementAccepted>{acceptLicenseAgreement}</agr:licenseAgreementAccepted>
-        </agr:licenseAgreement>
-        <agr:personalDataStatement>
-          <agr:signerId>{userId}</agr:signerId>
-          <agr:dateSigned>{date}</agr:dateSigned>
-          <agr:containsPrivacySensitiveData>{privacy}</agr:containsPrivacySensitiveData>
-        </agr:personalDataStatement>
-      </agr:agreements>
   }
 }
 
@@ -112,30 +77,9 @@ object DatasetMetadata {
   def apply(input: JsonInput): Try[DatasetMetadata] = input.deserialize[DatasetMetadata]
 
   private val doiScheme = "id-type:DOI"
-  private val dateScheme = "dcterms:W3CDTF"
 
-  type Date = QualifiedSchemedValue[String, DateQualifier]
-
-  def Date(value: DateTime,
-           qualifier: DateQualifier
-          ): Date = {
-    QualifiedSchemedValue[String, DateQualifier](
-      Some(dateScheme),
-      value.toString(ISODateTimeFormat.date()),
-      qualifier
-    )
-  }
-
-  private def missingValue(label: String) = {
+  def missingValue(label: String): InvalidDocumentException = {
     InvalidDocumentException(s"Please set $label in DatasetMetadata")
-  }
-
-  private def toBoolean(privacySensitiveDataPresent: PrivacySensitiveDataPresent): Try[Boolean] = {
-    privacySensitiveDataPresent match {
-      case PrivacySensitiveDataPresent.yes => Success(true)
-      case PrivacySensitiveDataPresent.no => Success(false)
-      case PrivacySensitiveDataPresent.unspecified => Failure(missingValue("PrivacySensitiveDataPresent"))
-    }
   }
 
   object PrivacySensitiveDataPresent extends Enumeration {
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
@@ -1,0 +1,16 @@
+package nl.knaw.dans.easy.deposit.docs
+
+import org.joda.time.DateTime
+
+import scala.util.Try
+import scala.xml.Elem
+
+object DatasetXml {
+  def apply(dateSubmitted: DateTime, dm: DatasetMetadata): Try[Elem] = {
+    for {
+      _ <- dm.licenceAccepted
+      privacy <- dm.privacyBoolean
+    } yield
+      <stub/>
+  }
+}
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
@@ -1,69 +1,121 @@
 package nl.knaw.dans.easy.deposit.docs
 
-import nl.knaw.dans.easy.deposit.docs.DatasetMetadata.DateQualifier.{ available, created }
-import nl.knaw.dans.easy.deposit.docs.DatasetMetadata.{ AccessRights, Author, QualifiedSchemedValue, SchemedKeyValue }
+import nl.knaw.dans.easy.deposit.docs.DatasetMetadata.DateQualifier.{ DateQualifier, available, created }
+import nl.knaw.dans.easy.deposit.docs.DatasetMetadata._
 
 import scala.util.Try
 import scala.xml.Elem
 
 object DatasetXml {
-  def apply(dm: DatasetMetadata): Try[Elem] = Try {
-    // TODO dc* namespaces
+  private type Lang = Option[SchemedKeyValue[String]]
+  private val otherDates = DateQualifier.values.filter(d => d != created && d != available).toSeq
+
+  def apply(dm: DatasetMetadata): Try[Elem] = Try { // TODO "dct:" or "dcterms:"
     <ddm:DDM
       xmlns:dc="http://purl.org/dc/elements/1.1/"
-      xmlns:ddm="http://easy.dans.knaw.nl/schemas/md/ddm/"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:dct="http://purl.org/dc/terms/"
+      xmlns:dcx-dai="http://easy.dans.knaw.nl/schemas/dcx/dai/"
+      xmlns:ddm="http://easy.dans.knaw.nl/schemas/md/ddm/"
+      xsi:schemaLocation="http://easy.dans.knaw.nl/schemas/md/ddm/ http://easy.dans.knaw.nl/schemas/md/2017/09/ddm.xsd"
     >
       <ddm:profile>
-        { requiredElems(dm.titles, "dc:title") }
-        { requiredElems(dm.descriptions, "dc:description") }
+        { requiredElems(dm.titles, "dct:title", dm.languageOfDescription) }
+        { requiredElems(dm.alternativeTitles, "dct:alternative", dm.languageOfDescription) }
+        { requiredElems(dm.descriptions, "dc:description", dm.languageOfDescription) }
         { requiredElems(dm.creators, "dcx-dai:creatorDetails") }
-        { requiredElems(dm.dates.map(filter(_, created)), "ddm:created") }
-        { requiredElems(dm.dates.map(filter(_, available)), "ddm:available") }
+        { requiredElems(dm.dates.map(filter(_, Seq(created))), "ddm:created") }
+        { requiredElems(dm.dates.map(filter(_, Seq(available))), "ddm:available") }
         { requiredElems(dm.audiences, "ddm:audience") }
         { requiredElems(dm.accessRights.map(Seq(_)), "ddm:accessRights") }
       </ddm:profile>
+      <ddm:dcmiMetadata>
+        { elems(dm.contributors, "dcx-dai:creatorDetails") }
+        { elems(dm.dates.map(filter(_, otherDates)), "otherDate") }
+        { elems(dm.subjects, "ddm:subject") }
+        { elems(dm.relations, "ddm:relation", dm.languageOfDescription) }
+        { optionalElem(dm.license, "dct:license") /* TODO xsi:type="dct:URI" not supported by json */ }
+      </ddm:dcmiMetadata>
+      <ddm:additional-xml>
+      </ddm:additional-xml>
     </ddm:DDM>
   }
 
-  private def elem[T](target: String)(source: T): Elem = (source match {
-    case QualifiedSchemedValue(Some(scheme), value, _) => <key scheme={scheme.toString}>{value}</key>
-    case QualifiedSchemedValue(None, value, _) => <key>{value}</key> // TODO so far just date-created/available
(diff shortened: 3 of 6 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
```diff
@@ -61,83 +58,31 @@ case class DatasetMetadata(identifiers: Option[Seq[SchemedValue[String]]] = None
...
-  private lazy val submitDate: Option[String] = {
-    dates.flatMap(_.find(_.qualifier == DateQualifier.dateSubmitted)).map(_.value)
+  lazy val privacyBoolean: Try[Boolean] = privacySensitiveDataPresent match {
+    case PrivacySensitiveDataPresent.yes => Success(true)
+    case PrivacySensitiveDataPresent.no => Success(false)
+    case PrivacySensitiveDataPresent.unspecified => Failure(missingValue("PrivacySensitiveDataPresent"))
   }
 
-  lazy val xml: Try[Elem] = Success(<stub/>) // TODO
+  lazy val licenceAccepted: Try[Unit] = if (acceptLicenseAgreement) Success(())
```
rvanheest: great job! I like this style! Would it make sense to actually let this be the entry point of `acceptLicenseAgreement` and make that field private? Now you have two entry points to that field...

src/main/scala/nl.knaw.dans.easy.deposit/docs/AgreementsXml.scala
```diff
@@ -0,0 +1,46 @@
...
+ */
+package nl.knaw.dans.easy.deposit.docs
+
+import org.joda.time.DateTime
+
+import scala.util.{ Failure, Success, Try }
+import scala.xml.Elem
+
+object AgreementsXml {
+  def apply(userId: String, dateSubmitted: DateTime, dm: DatasetMetadata): Try[Elem] = {
```
rvanheest: 👍 for this pattern

@alexal14 can you also do a similar style for `files.xml`?

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
```diff
@@ -0,0 +1,132 @@
...
+      xmlns:dc="http://purl.org/dc/elements/1.1/"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:dcterms="http://purl.org/dc/terms/"
+      xmlns:dcx-dai="http://easy.dans.knaw.nl/schemas/dcx/dai/"
+      xmlns:ddm="http://easy.dans.knaw.nl/schemas/md/ddm/"
+      xsi:schemaLocation="http://easy.dans.knaw.nl/schemas/md/ddm/ http://easy.dans.knaw.nl/schemas/md/2017/09/ddm.xsd"
+    >
+      <ddm:profile>
+        { requiredElems(dm.titles, "dcterms:title").addAttr(lang) }
+        { elems(dm.alternativeTitles, "dcterms:alternative").addAttr(lang) }
```
rvanheest: comparing with [easy-split-multi-deposit](https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/ea7c2dc3d6a78cacd779d2efcb2d799e399e22cf/src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddDatasetMetadataToDeposit.scala#L261): there `dcterms:alternative` goes into the `dcmiMetadata` part.

Not sure if it actually can go here!

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
```diff
@@ -0,0 +1,132 @@
...
+  private def elemLabel[T](target: String, source: T) = {
+    val rightsholder = "rightsholder"
+    source match {
+      case x: QualifiedSchemedValue[_, _] if target == qualifier => x.qualifier.toString
+      case Author(_, _, _, _, Some(SchemedKeyValue(_, _, `rightsholder`)), _, _) => "dcterms:rightsholder"
+      case _ => target
+    }
+  }
+
+  private def authorDetails(author: Author, lang: MaybeAttribute) =
```
rvanheest: when the author is an organization and it has a role, the output would be like:

https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/ea7c2dc3d6a78cacd779d2efcb2d799e399e22cf/src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddDatasetMetadataToDeposit.scala#L88-L89

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
```diff
@@ -0,0 +1,132 @@
...
+import nl.knaw.dans.easy.deposit.docs.DatasetMetadata.DateQualifier.{ available, created }
+import nl.knaw.dans.easy.deposit.docs.DatasetMetadata._
+
+import scala.util.Try
+import scala.xml.{ Attribute, Elem, Null, PrefixedAttribute }
+
+object DatasetXml {
+  private type MaybeAttribute = Option[Attribute]
+
+  private val otherDates = DateQualifier.values.filter(d => d != created && d != available).toSeq
```
rvanheest: `created` -> `DateQualifier.created`, analogous for `available`
would that make it clearer?

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
```diff
@@ -0,0 +1,132 @@
...
+        { requiredElems(dm.dates.map(filter(_, Seq(available))), "ddm:available") }
+        { requiredElems(dm.audiences, "ddm:audience") }
+        { requiredElems(dm.accessRights.map(Seq(_)), "ddm:accessRights") }
+      </ddm:profile>
+      <ddm:dcmiMetadata>
+        { elems(dm.contributors, "dcx-dai:creatorDetails", lang) }
+        { elems(dm.publishers, "dcterms:publisher").addAttr(lang) }
+        { elems(dm.sources, "dc:source").addAttr(lang) }
+        { elems(dm.dates.map(filter(_, otherDates)), qualifier) }
+        { optionalElem(dm.license, "dcterms:license") /* TODO xsi:type="dcterms:URI" not supported by json */ }
```
rvanheest: I assume you will continue this? I'm missing some fields...

General remark about this `dataset.xml`:
Have you looked at [AddDatasetMetadataToDeposit.scala](https://github.com/rvanheest-DANS-KNAW/easy-split-multi-deposit/blob/ea7c2dc3d6a78cacd779d2efcb2d799e399e22cf/src/main/scala/nl.knaw.dans.easy.multideposit/actions/AddDatasetMetadataToDeposit.scala#L279-L278) in easy-split-multi-deposit. Not about the coding style (I like this one way more!), but about the content. Can we make sure that both applications produce similar DDM with the same tags, in the same order, etc.?

For example: you're using `dcterms:titles` while SMD is using `dc:titles`; you're using `dc:description` while SMD is using `dcterms:description`, etc.

src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
```diff
@@ -181,19 +182,17 @@ class DepositDirSpec extends TestSupportFixture with MockFactory {
   "writeSplittedDatasetMetadata" should "write 4 files" in {
     val prologue = """<?xml version='1.0' encoding='UTF-8'?>"""
     val message = "Lorum ipsum"
-    val datasetMetadata = DatasetMetadata(
-      privacySensitiveDataPresent = PrivacySensitiveDataPresent.yes,
-      acceptLicenseAgreement = true,
-      messageForDataManager = Some(message)
-    )
+    val datasetMetadata = DatasetMetadata((File("src") / "test" / "resources" / "manual-test" / "datasetmetadata-from-ui-all.json").contentAsString)
```
rvanheest: `File("src") / "test" / "resources" / "manual-test" / "datasetmetadata-from-ui-all.json"`
probably the same as 
`File(getClass.getResource("/manual-test/datasetmetadata-from-ui-all.json"))`

(or maybe without the `/` prefix)

same in other places in your tests

rvanheest: btw. what manual tests are there...
shouldn't you migrate those to automated tests?

src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
```diff
@@ -181,19 +182,17 @@ class DepositDirSpec extends TestSupportFixture with MockFactory {
...
     val deposit = createDepositAsPreparation("user001")
     val mdDir = deposit.getDataFiles.getOrElse(fail("preconditions are not met"))
       .filesMetaData.parent.createIfNotExists(asDirectory = true, createParents = true)
     deposit.writeDatasetMetadataJson(datasetMetadata)
     val oldSize = (mdDir / "dataset.json").size
 
-    deposit.splitDatasetMetadata shouldBe Success(())
-    (mdDir / "dataset.json").size should be > oldSize
+    deposit.splitDatasetMetadata(DateTime.now) shouldBe Success(())
+    (mdDir / "dataset.json").size shouldBe oldSize // date submitted is no longer addededd
```
rvanheest: `addededd` --> `added`?

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -0,0 +1,153 @@
...
+      |  "audiences": [ { "scheme": "", "key": "", "value": ""} ]
+      |}""".stripMargin).getOrElse(fail("parsing minimal json failed"))
+
+  "apply" should "produce DDM from  minimal json" in {
+    DatasetXml(minimal) shouldBe a[Success[_]]
+  }
+
+  it should "report a missing title" in {
+    DatasetXml(minimal.copy(titles = None)) should matchPattern {
+      case Failure(e) if e.getMessage == "no content for mandatory dcterms:title" =>
```
rvanheest: 👍 for this `case ... if ... =>` syntax
I don't think I've ever seen that before, but that's really nice. Probably also gives a better error message than using `inside` for this.

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
jo-pol: Would json4s deserialize private fields?

rvanheest: note sure, but feel free to test it 😉

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
jo-pol: I was triggered by https://github.com/DANS-KNAW/easy-schema/blob/master/src/main/assembly/dist/docs/examples/ddm/example2.xml#L63 

As the check boxes tell, I intend to validate against the DDM schema and still have to check the spec-matrix line by line

rvanheest: awesome. Maybe you can use a validation similar to what was used by @janvanmansum in easy-validate-dans-bag?

src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
jo-pol: These files can be used for `PUT /deposit/{id}/metadata`.

Manual tests should also be tested automatically to make sure they stay valid.

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
jo-pol:
```scala
  private val otherDateQualifiers = DateQualifier.values.filter(qualifier => 
    qualifier != DateQualifier.created && qualifier != DateQualifier.available
  ).toSeq
```

rvanheest: can you use `filterNot` here? maybe that would better signify what you're doing?

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -0,0 +1,202 @@
...
+
+    val datasetMetadata = DatasetMetadata(readTestData("datasetmetadata-from-ui-all.json"))
+      .getOrElse(fail("could not parse test data"))
+    val actual = DatasetXml(datasetMetadata).getOrElse(fail("conversion to DDM failed"))
+
+    prettyPrinter.format(actual) shouldBe prettyPrinter.format(expected)
+  }
+
+  "DDM validation" should "succeed for the minimal json" in {
+    assume(triedSchema.isSuccess)
```
jo-pol: Note the variation on the usual [assume(canConnect)](https://github.com/DANS-KNAW/easy-solr4files-index/blob/f4f3c59cf212b83a037422ecac22d865beee3689/src/test/scala/nl.knaw.dans.easy.solr4files/TestSupportFixture.scala#L83-L95). 

In this case the **test** is performing the schema-validation, in the linked case the **tested code** does the validation. In this case we don't have the custom time out and the ignored test will take significantly longer when working offline. But adding the `canConnect` code would add an extra second when the schema's are accessible.

This might be the place to start experimenting with local copies of the full tree of schema's to get less dependent on external services, but that is beyond the scope of this issue.

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -22,22 +22,73 @@ import better.files.File
...
+
+  // pretty provides friendlier trouble shooting for complex XML's than Utility.trim
+  private val prettyPrinter: PrettyPrinter = new scala.xml.PrettyPrinter(1024, 2)
+
+  /**
+   * @param input          a Success(DatasetMetadata)
+   * @param expectedOutput <ddm:profile> and optionaly <ddm:dcmiMetadata>
+   *                       if provided, input should convert into a DDM with the same content
+   */
+  def validDatasetMetadata(input: Try[DatasetMetadata], expectedOutput: Seq[Node]): Unit = {
```
rvanheest: since `expectedOutput` is `Seq.empty` so many times, why don't you add that as the default parameter?

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -22,22 +22,73 @@ import better.files.File
...
+    lazy val datasetMetadata = input.recoverWith { case e =>
```
rvanheest: `recoverWith` just to do a side-effect is called `doIfFailure` (see dans-scala-lib)

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -22,22 +22,73 @@ import better.files.File
...
+      Failure(e)
+    }.getOrElse(fail("can't create DDM from test input"))
+
+    it should "generate expected DDM" in {
+      // TODO change into "should contain": leafs in expected should occur with same path in ddm
+      // perhaps with something similar to https://stackoverflow.com/questions/39001421/
+      // or a third argument (with xpaths?) to extract subsets from the generated DDM
+      val expected = emptyDDM.copy(child = expectedOutput)
+      if(expectedOutput.nonEmpty)
+        prettyPrinter.format(ddm) shouldBe
```
rvanheest: in [easy-split-multi-deposit](https://github.com/rvanheest-DANS-KNAW/easy-split-multi-deposit/blob/ea7c2dc3d6a78cacd779d2efcb2d799e399e22cf/src/test/scala/nl.knaw.dans.easy.multideposit/CustomMatchers.scala#L41) I made a `should ...` operator for comparing XMLs. Maybe you can use that one here as well?

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -22,22 +22,73 @@ import better.files.File
...
 import scala.util.{ Failure, Success, Try }
-import scala.xml.{ Node, Utility }
+import scala.xml.{ Elem, Node, PrettyPrinter, Utility }
 
-class DatasetXmlSpec extends TestSupportFixture {
+trait DdmBehavior {
+  this: TestSupportFixture =>
 
-  private lazy val triedSchema = Try { // loading postponed until we actually start validating
+  lazy val triedSchema: Try[Schema] = Try { // loading postponed until we actually start validating
```
rvanheest: you're doing something with `getOrElse` and `fail` below. Would it make sense to apply that here as well?

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
jo-pol: I would not like to loose the IntelliJ supported diff, nor have to wade through big one liners for trouble shooting.

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -0,0 +1,296 @@
...
+                          ): Unit = {
+    lazy val datasetMetadata = input
+      .doIfFailure{ case e => println(s"$e")}
+      .getOrElse(fail("test input should be a success"))
+    lazy val ddm = DatasetXml(datasetMetadata)
+      .doIfFailure{ case e => println(s"$e")}
+      .getOrElse(fail("can't create DDM from test input"))
+
+    it should "generate expected DDM" in {
+      if (expectedOutput.nonEmpty)
```
jo-pol: @rvanheest 
Not completely satisfied here. Tried `assume` too, but that gives a worse wrong impression. But now it suggests the output is as expected while it is not asserted at all.

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
rvanheest: https://github.com/DANS-KNAW/easy-split-multi-deposit/pull/113

I'll also copy this to easy-bag-index

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
```diff
@@ -221,7 +198,7 @@ object DatasetMetadata {
                       url: Option[String],
                       title: Option[String],
                      ) extends RelationType {
-    require(isValid, "Relation needs at least one of (title | url) got: ${toJson(this)}")
+    require(isValid, s"Relation needs at least one of (title | url) got: ${toJson(this)}")
 
     def isValid: Boolean = {
```
rvanheest: this method is only used in the `require` above. Are you planning to use it in other places as well? Otherwise inlining this check wouldn't be a bad idea, IMO.

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
```diff
@@ -0,0 +1,171 @@
...
+        s"no content for mandatory $target"
+      ))
+    }.map(elem(target, lang))
+  }
+
+  private def isRightsHolder[T](a: T with Author) = {
+    a.role.exists(_.key == "RightsHolder")
+  }
+
+  private def keepNonEmpty[T](ts: Seq[T]) = {
```
rvanheest: shouldn't this also filter out the `Option.empty` cases?

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
```diff
@@ -0,0 +1,171 @@
...
+
+  private def requiredElems[T](source: Option[Seq[T]], target: String, lang: Option[Attribute] = None): Seq[Elem] = {
+    source.map(keepNonEmpty).filterNot(_.isEmpty).getOrElse{
+      throw InvalidDocumentException("DatasetMetadata",  new Exception(
+        s"no content for mandatory $target"
+      ))
+    }.map(elem(target, lang))
+  }
+
+  private def isRightsHolder[T](a: T with Author) = {
```
rvanheest: why `T with Author` and not just `Author`?

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
```diff
@@ -0,0 +1,171 @@
...
+      created, // in ddm:profile
+      available, // in ddm:profile
+      dateSubmitted // generated, ignore if in input
+    ).contains(qualifier)
+  }.toSeq
+
+  // TODO make private together with implicit class RichElem
+  val targetFromQualifier = "qualifier"
+
+  def apply(dm: DatasetMetadata): Try[Elem] = Try {
```
rvanheest: I have to do a lot of mental work to grasp and understand what is exactly going on here and what the result will be. This is especially the case with recursive use of methods like `requiredElems` such as in [`orgDetails`](https://github.com/DANS-KNAW/easy-deposit-api/pull/54/files#diff-d0643e927258e9fc26054ff4e75b54c8R105). Given that you have to wrap it in `Some(Seq(...))` feels a bit weird.

I'm sure you have the best intentions with this class, but to me it feels a bit overdesigned. Is there any change we can simplify stuff somewhat more?

src/test/resources/manual-test/readme.md
```diff
@@ -0,0 +1,14 @@
+JSON files for manual tests
+---------------------------
+
+* Launch a virtual machine with `vagrant up` in the root of the project
+* Create a deposit, for details see `docs/api.html`
+* Use the returned UUID to replace XXX in the following command.
+
+    curl -i  -H 'Content-Type: text/plain'  --data-binary "@XXX.json" -X POST -u user001:user001 'http://test.dans.knaw.nl:20190/deposit/XXX/datasetmetadata'
```
rvanheest: Maybe you can add some more things here about:

1. what should the outcome of this call be?
2. what are we testing here/what is the purpose of this test?
3. you would do future you or other developers a nice service if you would spell out the commands for creating a deposit instead of pointing them to the documentation.
4. make a link to the documentation instead.

rvanheest: unrelated to previous comment: what does this `@` do in `"@XXX.json"`

src/test/resources/manual-test/readme.md
```diff
@@ -0,0 +1,14 @@
...
+
+
+Purpose of HTML files
+---------------------
+
+Examine cookies with developer tools of a browser without a fully dressed web client.
```
rvanheest: A couple more words on the following things would be helpful:
1. the purposes of each individual HTML file
2. what you can test with them?
3. what steps do you want me to take while testing?
4. what you should expect to find with them?

_future you thanks you on behalf of me and all other developers_ 😉

src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
rvanheest: friendly reminder for the first comment...

src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
```diff
@@ -181,19 +182,17 @@ class DepositDirSpec extends TestSupportFixture with MockFactory {
...
     val deposit = createDepositAsPreparation("user001")
     val mdDir = deposit.getDataFiles.getOrElse(fail("preconditions are not met"))
       .filesMetaData.parent.createIfNotExists(asDirectory = true, createParents = true)
     deposit.writeDatasetMetadataJson(datasetMetadata)
     val oldSize = (mdDir / "dataset.json").size
 
-    deposit.splitDatasetMetadata shouldBe Success(())
-    (mdDir / "dataset.json").size should be > oldSize
+    deposit.splitDatasetMetadata(DateTime.now) shouldBe Success(())
+    (mdDir / "dataset.json").size shouldBe oldSize // date submitted is no longer added
```
rvanheest: Make this comment

`// while date submitted is added to the dataset metadata, it is not added to the dataset.json file`

The current comment defines the current change, which isn't relevant in the future anymore.

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -0,0 +1,328 @@
...
+        surname = Some("Terix"),
+        role = Some(SchemedKeyValue("", "RightsHolder", ""))
+      )
+    ))
+    validDatasetMetadata(
+      input = Success(minimal.copy(creators = someCreators, contributors = someContributors)),
+      subset = { actualDDM => emptyDDM.copy(child = actualDDM \ "dcmiMetadata") },
+      expectedOutput = Seq(
+        //N.B: creators in ddm:profile unless they are rightsHolders
+        <ddm:dcmiMetadata>
```
rvanheest: There are some formatting issues here with your XML.
Note that XML is sometimes not touched by our formatter.
Probably it helps to format it properly here and then add `// @formatter:off` and `// @formatter:on` before and after it respectively.

That formatting thingy is probably not a bad idea for the rest of the XML snippets in the code base as well.

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -0,0 +1,328 @@
...
+    input = parseTestResource("datasetmetadata.json")
+  )
+  // TODO keep resources in sync with UI module: https://github.com/DANS-KNAW/easy-deposit-ui/blob/784fdc5/src/test/typescript/mockserver/metadata.ts#L246
+  "datasetmetadata-from-ui-some.json" should behave like validDatasetMetadata(
+    input = parseTestResource("datasetmetadata-from-ui-some.json")
+  )
+  "datasetmetadata-from-ui-all.json without one of the authors" should behave like validDatasetMetadata(
+    input = parseTestResource("datasetmetadata-from-ui-all.json"),
+    expectedOutput = Seq(
+      <ddm:profile>
```
rvanheest: would it make sense to extract the larger snippets in this class in either their own values or even their own files?

jo-pol: I started with this one a kind test driven development of the implementation. In the next iteration I intend to reduce this one to just validate the DDM. Then all variations of individual fields should have smaller tests like the authors and dates.

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
jo-pol: Where you say 'somteimes' I experience 'never'

It would be great to have formatting variants. When just XML is selected, format it; if more or nothing is selected, don't format XML.

src/test/resources/manual-test/readme.md
jo-pol: see man pages of curl

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
rvanheest: if you find something in the IntelliJ formatter, feel free to do a pull request!

src/main/scala/nl.knaw.dans.easy.deposit/docs/AgreementsXml.scala
```diff
@@ -0,0 +1,46 @@
...
+    for {
+      _ <- dm.licenceAccepted
+      privacy <- dm.privacyBoolean
+    } yield
+      <agr:agreements
```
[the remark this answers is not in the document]
jo-pol: To prevent too many arguments (or repeat the argument list with more element) I'm temped to create nested def's. Usually you don't like that too.

src/test/resources/manual-test/readme.md
rvanheest: 👍
> If you start the data with the letter @, the rest should be a filename.

src/main/scala/nl.knaw.dans.easy.deposit/docs/AgreementsXml.scala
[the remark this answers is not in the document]
jo-pol: The following solution looks even more complex if we would have more tried values.

```scala
    val privacy = for {
      _ <- dm.licenceAccepted
      privacy <- dm.privacyBoolean
    } yield privacy
    
    privacy.map(privacy =>...)
```

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
jo-pol: I got an headache figuring out the line that calls this. At that line I want a `None` when I have an `Option[Seq[String]]` with just empty strings, otherwise the empty strings weeded out of the `Seq`.

Now that I'm trying to explain, the `elems` method should also weed out the empty strings, and the generic classes in DatasetMetadata should all have a `require` that rejects mandatory fields with empty strings.

rvanheest: May I suggest that (given that you got a headache from figuring out just a single line in this class) this is a sign that the `DatasetXml` class is probably over-engineered. Although you found a nice pattern, I think it is not very maintainable in this state, unless it will be simplified further. Given that you're currently doing just the 'simple' elements, I'm starting to worry how much effort it's going to cost you and/or other developers to add more complex elements in the future. Perhaps you should first focus on getting the XML generated in a simple way and refactor it later on. As you mentioned to me many times before:

> Premature optimization is the root of all evil.

image

src/main/scala/nl.knaw.dans.easy.deposit/docs/AgreementsXml.scala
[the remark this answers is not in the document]
rvanheest: Not sure if that's what I meant. First of all, I would say that you're doing both validation and creating XML in the same place. I think, just like `DatasetXml`, this method should only be concerned with creating the XML, not with the validation part. If that validation fails, then this method should not be called.

Second, what I meant with extract method is to create something like:

```scala
private def createXml(userId: String, dateSubmitted: DateTime, acceptLicenseAgreement: Boolean, privacy: Boolean): Elem
```

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetXmlSpec.scala
```diff
@@ -285,7 +278,15 @@ class DatasetXmlSpec extends TestSupportFixture with DdmBehavior {
...
-    )
+  )
+
+  private def dcmiMetadata(actualDDM: Elem) = emptyDDM.copy(child =
+    actualDDM \ "dcmiMetadata"
+  )
+
+  private def profileElements(actualDDM: Elem, label: String) = emptyDDM.copy(child =
+    // prevent namespaces on children
+    Elem("ddm", "profile", Null, emptyDDM.scope, minimizeEmpty = true, actualDDM \ "profile" \ label: _*)
```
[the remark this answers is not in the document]
jo-pol: Same as above. Or perhaps worse because items would be returned at the wrong level.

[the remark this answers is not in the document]
rvanheest: 👍

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetXml.scala
```diff
@@ -0,0 +1,172 @@
...
+
+  private def orgDetails(role: Option[SchemedKeyValue[String]] = None)
+                        (organization: String)
+                        (implicit lang: Option[Attribute]) =
+      <dcx-dai:organization>
+        { role.toSeq.map(role => <dcx-dai:role>{ role.key }</dcx-dai:role>) }
+        { <dcx-dai:name>{ organization }</dcx-dai:name>.addAttr(lang) }
+      </dcx-dai:organization>
+
+  private implicit class SubmittedDatasetMetadata(dm: DatasetMetadata) extends DatasetMetadata {
```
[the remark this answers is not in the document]
jo-pol: The recalculations preparing for the conversion started with some simple inline calls to generic implicit methods like the current getNonEmpty, but the number of lines I needed to extract into a value grew and grew. I wanted to clean up the XML conversion method by grouping the growing number of values that needed recalculation. 

We need a case class during the draft stage for json conversions so we can't extend the class for the submit stage. Adding many submit values to the base class would cause needles calculations during the draft stage, but most of all we should not apply submit requirements during the draft stage. At that point the idea for a complex implicit class was born.

For now I could reduce the original clutter for the apply method into rightsHolders and SubmittedDates. Don't yet know what will emerge in the next implementation phase with temporal/spatial coverage and so on.

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
```diff
@@ -183,75 +139,131 @@ object DatasetMetadata {
...
   case class PossiblySchemedKeyValue[S](scheme: Option[S],
                                         key: Option[String],
                                         value: String,
-                                       )
+                                       ) extends RequiresNonEmpty {
+    requireNonEmptyString(value, "value")
+    val hasScheme: Boolean = scheme.isProvided
+    val hasKey: Boolean = key.isProvided
+  }
 }
```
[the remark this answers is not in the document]
jo-pol: I called the package `docs` with the intention to contain classes/documents stored in files or exchanged with the UI. Perhaps the JsonUtil should go into a package object to be consistent and these members should go into a `dm` sub-package?

PR #59, opened 2018-07-31
EASY-1564 deposit properties + EASY-1464 submit: doi and state

jo-pol:
Fixes EASY-1464 + some more steps of EASY-1464

#### When applied it will
* add the fields that wre still missing from the deposit.properties
* add a DOI if not yet available
* check state transition before submit actions
* change the state

SubmitterSpec and a new test in IntegrationSpec illustrate the implemented changes.

Submit will fail with a not implemented exception but metadata files should be created/changed.

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
@@ -166,11 +166,13 @@ case class DepositDir private(baseDir: File, user: String, id: UUID) extends Deb
    * Writes the dataset level metadata for this deposit.
    *
    * @param md the metadata to write
-   *           Changing "identifiers": [{ "scheme": "id-type:DOI", "value": "..."}]
-   *           will cause an error when calling getDOI by an explicit request
-   *           or when getDOI is called implicitly by submit
-   *           because it will not match the deposit properties.
-   *           A "dates": [ { ..., "qualifier": "dcterms:dateSubmitted" }]
+   *           In terms of JSon syntax, content like
+   *           "identifiers": [{ "scheme": "id-type:DOI", "value": "..."}]
+   *           should have been acquired with an explicit getDOI request.
+   *           Otherwise submit will fail because the value
+   *           is out of sync with deposit properties.
+   *           JSon content like
+   *           "dates": [ { ..., "qualifier": "dcterms:dateSubmitted" }]
    *           will cause an error when converted to dataset.xml at submit.
    */
   def writeDatasetMetadataJson(md: DatasetMetadata): Try[Unit] = Try {
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
@@ -157,15 +157,20 @@ case class DepositDir private(baseDir: File, user: String, id: UUID) extends Deb
    * Writes the dataset level metadata for this deposit.
    *
    * @param md the metadata to write
+   *           Changing "identifiers": [{ "scheme": "id-type:DOI", "value": "..."}]
+   *           will cause an error when calling getDOI by an explicit request
+   *           or when getDOI is called implicitly by submit
+   *           because it will not match the deposit properties.
+   *           A "dates": [ { ..., "qualifier": "dcterms:dateSubmitted" }]
+   *           will cause an error when converted to dataset.xml at submit.
    */
   def writeDatasetMetadataJson(md: DatasetMetadata): Try[Unit] = Try {
-    // TODO DOI should not change, only enforced by client calling getDOI. State and dateSubmitted should not change either.
     datasetMetadataJsonFile.write(toJson(md))
     () // satisfy the compiler which doesn't want a File
   }.recoverWith { case _: NoSuchFileException => notFoundFailure() }
 
   /** create dataset.xml, agreements.xml, files.xml
-   *  from datasetmetadata.json and files in data folder
+   * from datasetmetadata.json and files in data folder
    */
   def createXMLs(dateSubmitted: DateTime): Try[Unit] = {
     for {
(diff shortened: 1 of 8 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -282,7 +282,9 @@ object DepositDir {
       addProperty("creation.timestamp", depositInfo.date)
       addProperty("state.label", depositInfo.state.toString)
       addProperty("state.description", depositInfo.stateDescription)
-      addProperty("depositor.userId", user)
+      addProperty("curation.required", "yes")
```
lindareijnhoudt: why has the depositor.userId disappeared?

src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -75,21 +75,23 @@ case class DepositDir private(baseDir: File, user: String, id: UUID) extends Deb
    */
   def setStateInfo(stateInfo: StateInfo): Try[Unit] = {
     for {
-      props <- getDepositProps
-      currentState = State.withName(props.getString("state.label"))
-      _ <- checkStateTransition(currentState, stateInfo.state)
+      props <- checkStateTransition(stateInfo.state) // read again as submit might have added a DOI
```
lindareijnhoudt: isn't it a bit weird to get the props returned for this check? I liked the previous def better, where there were two state-parameters

janvanmansum: I agree, it is almost always best [not to mix queries and commands](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation).

src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -157,15 +159,20 @@ case class DepositDir private(baseDir: File, user: String, id: UUID) extends Deb
    * Writes the dataset level metadata for this deposit.
    *
    * @param md the metadata to write
+   *           Changing "identifiers": [{ "scheme": "id-type:DOI", "value": "..."}]
```
janvanmansum: @jo-pol : could your rephrase this, so that I can understand it?

jo-pol:
I tend to avoid the boiler plate of custom exceptions for internal errors that don't need handling and are presented to the outside world as an internal server error. These exceptions just provide more informative messages in the log. Perhaps it is an idea to create one custom exception for those cases.

janvanmansum:
@jo-pol : they don't always have to be *custom* exceptions, you can often find JRE exception that covers it. I use `IllegalArgumentException` a lot for invalid input and `IllegalStateException` for when the application is in a corrupt state for example.

I still see some use of the `Exception` base class. I strongly suggest to refactor that in a later PR.

Also in `nl.knaw.dans.easy.deposit.DepositDir#notFoundFailure` you create an exception as the "cause" of another one. This might also get confusiong. Why not pass in the exception that you catch in `nl.knaw.dans.easy.deposit.DepositDir#getDatasetMetadata`?

src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -157,15 +166,22 @@ case class DepositDir private(baseDir: File, user: String, id: UUID) extends Deb
    * Writes the dataset level metadata for this deposit.
    *
    * @param md the metadata to write
+   *           In terms of JSon syntax, content like
```
rvanheest: JSon -> JSON?

src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -276,6 +292,9 @@ object DepositDir {
       addProperty("state.label", depositInfo.state.toString)
       addProperty("state.description", depositInfo.stateDescription)
       addProperty("depositor.userId", user)
+      addProperty("curation.required", "yes")
```
rvanheest: not sure (1) if you know this and (2) if this is necessary here, but there is a `BooleanUtils.toStringYesNo` method for the `PropertiesConfiguration` from Apache Commons.

src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
```diff
@@ -161,9 +160,7 @@ class DepositDirSpec extends TestSupportFixture with MockFactory {
     val deposit = createDepositAsPreparation(user)
     (deposit.baseDir / user / deposit.id.toString / "bag" / "metadata" / "dataset.json").writeText(s"""{"doi":"$doi"}""")
 
-    val pidMocker = mock[PidRequester] // note that no pid is requested
-
-    deposit.getDOI(pidMocker) should matchPattern { case Failure(CorruptDepositException(_, _, _)) => }
+    deposit.getDOI(null) should matchPattern { case Failure(CorruptDepositException(_, _, _)) => }
   }
 
   it should """return the available DOI""" in {
```
rvanheest: why all these triple quotes?

src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
```diff
@@ -161,9 +160,7 @@ class DepositDirSpec extends TestSupportFixture with MockFactory {
     val deposit = createDepositAsPreparation(user)
     (deposit.baseDir / user / deposit.id.toString / "bag" / "metadata" / "dataset.json").writeText(s"""{"doi":"$doi"}""")
 
-    val pidMocker = mock[PidRequester] // note that no pid is requested
-
-    deposit.getDOI(pidMocker) should matchPattern { case Failure(CorruptDepositException(_, _, _)) => }
+    deposit.getDOI(null) should matchPattern { case Failure(CorruptDepositException(_, _, _)) => }
```
rvanheest: what's this `null` doing here?

PR #61, opened 2018-08-09
Specified status code 400 for when an invalid ZIP is sent.

janvanmansum:
Proposed addition to API.


rvanheest (approved):

lindareijnhoudt:
do you still expect a `Content-Disposition` header when a zip-file is sent?
What would be the reponse if this header is omitted?

jo-pol (approved):

PR #62, opened 2018-08-13
EASY-1659 Implement GET file-list

jo-pol:
Fixes EASY-1659 Implement GET file-list

#### When applied it will
* maintain the draft payload with the bag-lib
* _not yet_ maintain dataset.json with the bag-lib
* because of the previous bullet the draft bag is not yet validated at submission

#### Related pull requests on github
repo                       | PR                | note
-------------------------- | ----------------- | ----
easy-                      | [PR#](PRlink)     | ..

PR #63, opened 2018-08-15
EASY-1657 post file using content-disposition header

jo-pol:
Fixes EASY-1657 Implement upload single file with POST directory

#### When applied it will
* interpret the specs as sub-requests of a request with `multipart/form-data`
* `Content-Type` of the sub-requests are still ignored, so zip-files [EASY-1658](https://drivenbydata.atlassian.net/browse/EASY-1658) are uploaded as-is.
* [x] unit tests
* The response code on success is 200-ok as it might be a mix of created and replaced. The specs says 201-created but that applies to subrequests. No specification how to wrap these up in the `multipart/form-data` request.

Some statistics:

* see the updated `test/resources/readme.md`
* test both PUT and POST
* test with curl: see this [stack-overflow](https://stackoverflow.com/questions/4238809/example-of-multipart-form-data)
* [x] large files https://github.com/DANS-KNAW/easy-deposit-api/pull/63#issuecomment-413111374

Uploading a file of 540 MB to `deasy` resulted in the browser complaining about "the connection was reset" and the log below (`HttpParser full`, replaced IP addresses with ...). After clearing the `bag/data` folder on the server I could upload it multiple times. Same for 1GB and even the 5GB succeeded on a second attempt. Might have to to with fooling around with the browser's back-button on the test page while changing the file to upload. It looks like a proper reset is not only clear `bag/data` but also POST a small file.

```
[2018-08-23 15:43:42,638] WARN  HttpParser Full for SCEP@747f9023{l(/...:59536)<->r(/...:20190),s=1,open=true,ishut=false,oshut=false,rb=false,wb=false,w=true,i=0r}-{AsyncHttpConnection@67754c1c,g=HttpGenerator{s=0,h=-1,b=-1,c=-1},p=HttpParser{s=-13,l=0,c=-3},r=1}
```
Some statistics of a successful upload of 5GB. Of course statistics with `teasy` would be more realistic that with `deasy`.

```
[2018-08-24 09:10:43,910] INFO  uploading to [/var/opt/dans.knaw.nl/deposit/drafts/user001/196fb09a-03c5-4951-a308-4879e4b7f2ab/bag] of [path/tox/large.txt]
[2018-08-24 09:12:37,632] INFO  created=true user001/196fb09a-03c5-4951-a308-4879e4b7f2ab/path/tox/large.txt
[2018-08-24 09:12:37,633] INFO  POST returned status=200 headers=Map()
-rw-r--r--. 1 easy-deposit-api easy-deposit-api 1663       2018-08-24 09:12:37.624723161 +0200 manifest-sha1.txt
-rw-r--r--. 1 easy-deposit-api easy-deposit-api 5007600000 2018-08-24 09:11:50.768755240 +0200 large.txt

upload:   10:43 tot 11:50 = 1:07
manifest: 11:50 tot 12:37 = 0:47
```

~~May cause merge conflicts on IntegrationSpec.java with #62~~ merged

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
@@ -133,7 +133,7 @@ class DepositServlet(app: EasyDepositApiApp) extends ProtectedServlet(app) {
           managedIS.apply(is => app.writeDepositFile(is, user.id, uuid, fullPath))
         case (_, _) =>
           val explanation = s"Expecting header 'Content-Type: $zip' or 'Content-Type: $bin'; the latter with a filename in the 'Content-Disposition'."
-          Failure(BadRequestException(s"$explanation GOT: $maybeContentType AND $maybeContentType"))
+          Failure(BadRequestException(s"$explanation GOT: $maybeContentType AND $mayBeFileName"))
       }
     } yield fileCreatedOrOkResponse(newFileWasCreated)
       ).getOrRecoverResponse(respond)
--- a/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
@@ -21,12 +21,12 @@ import java.util.UUID
 import nl.knaw.dans.easy.deposit.authentication.ServletEnhancedLogging._
 import nl.knaw.dans.easy.deposit.docs.JsonUtil.{ InvalidDocumentException, toJson }
 import nl.knaw.dans.easy.deposit.docs.{ DatasetMetadata, StateInfo }
-import nl.knaw.dans.easy.deposit.servlets.DepositServlet.InvalidResourceException
+import nl.knaw.dans.easy.deposit.servlets.DepositServlet.{ BadRequestException, InvalidResourceException }
 import nl.knaw.dans.easy.deposit.{ EasyDepositApiApp, _ }
 import org.scalatra._
 import resource.managed
 
-import scala.util.{ Failure, Try }
+import scala.util.{ Failure, Success, Try }
 
 class DepositServlet(app: EasyDepositApiApp) extends ProtectedServlet(app) {
 
@@ -87,17 +87,27 @@ class DepositServlet(app: EasyDepositApiApp) extends ProtectedServlet(app) {
       .map(depositFiles => Ok(body = toJson(depositFiles)))
       .getOrRecoverResponse(respond)
   }
-  post("/:uuid/file/*") { upload } //dir
-  put("/:uuid/file/*") { upload } //file
-  private def upload = {
-    for {
-      managedIS <- getRequestBodyAsManagedInputStream
-      newFileWasCreated <- managedIS.apply(is => forPath(app.writeDepositFile(is)))
-    } yield if (newFileWasCreated)
-              Created(headers = Map("Location" -> request.uri.toASCIIString))
-            else Ok()
-  }.getOrRecoverResponse(respond)
-
+  post("/:uuid/file/*") { //file
+    {
+      for { // a unique request signature so no forXxx method
+        managedIS <- getRequestBodyAsManagedInputStream
+        uuid <- getUUID
+        dir <- getPath
+        // TODO EASY-1658: mime type application/zip or application/binary
+        disposition <- getMandatoryHeader("Content-Disposition")
+        path = dir.resolve(disposition)
+        newFileWasCreated <- managedIS.apply(is => app.writeDepositFile(is)(user.id, uuid, path))
+      } yield newFileHeader(newFileWasCreated)
+    }.getOrRecoverResponse(respond)
+  }
+  put("/:uuid/file/*") { //file
+    {
+      for {
+        managedIS <- getRequestBodyAsManagedInputStream
+        newFileWasCreated <- managedIS.apply(is => forPath(app.writeDepositFile(is)))
+      } yield newFileHeader(newFileWasCreated)
+    }.getOrRecoverResponse(respond)
+  }
   delete("/:uuid/file/*") { //dir and file
     forPath(app.deleteDepositFile)
       .map(_ => NoContent())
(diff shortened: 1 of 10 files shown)


rvanheest:
I'm not entirely sure, but maybe http://scalatra.org/guides/2.6/formats/upload.html would be helpful to you when working on this issue.

Also, I'm not really sure if you're already finished with this issue, but have you already tried uploading some large files? (500 MB, couple GBs, etc.)

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -89,14 +89,19 @@ class DepositServlet(app: EasyDepositApiApp) extends ProtectedServlet(app) {
...
-        // TODO EASY-1658: mime type application/zip or application/binary
-        disposition <- getMandatoryHeader("Content-Disposition")
-        path = dir.resolve(disposition)
-        newFileWasCreated <- managedIS.apply(is => app.writeDepositFile(is)(user.id, uuid, path))
+        maybeContentDisposition = getLowerCaseHeaderValue("Content-Disposition")
+        maybeContentType = getLowerCaseHeaderValue("Content-Type")
+        newFileWasCreated <- (maybeContentType, maybeContentDisposition) match {
+          case (Some(`zip`), _) => Failure(???) // TODO issue EASY-1658
+          case (Some(`bin`), Some(contentDisposition: String)) =>
+            managedIS.apply(is => forPathSupplement(contentDisposition, app.writeDepositFile(is)))
```
rvanheest: this, as well as the implementation of `forPathSupplement`, looks more complicated than it should probably be.

1. I don't think you require a callback here
2. you're opening the `managedIS` using `apply` way too early. After all, `forPathSupplement` might already fail in the first 3 parts for the for-comprehension.
3. In the implementation of `forPathSupplement` I don't understand why you wrap the callback call in a `Try` and flatten it immediately. The comment there isn't clear either.

jo-pol: I tried to replace
https://github.com/DANS-KNAW/easy-deposit-api/blob/eb783743d1ac3359f394477c4cf26b2ce25acf96/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala#L34
https://github.com/DANS-KNAW/easy-deposit-api/blob/eb783743d1ac3359f394477c4cf26b2ce25acf96/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala#L39
with 
```
app.getDeposits(user.id)
app.createDeposit(user.id)
```
that broke a unit test with an unexpected exception because somewhere something was not 'honest'. So I decided to apply the same old pattern here. We might need to rethink that pattern.

src/test/scala/nl.knaw.dans.easy.deposit/servlets/IntegrationSpec.scala
```diff
@@ -194,9 +194,16 @@ class IntegrationSpec extends TestSupportFixture with ServletFixture with Scalat
 
     // upload a file
     expectsUserFooBar
-    post(uri = s"/deposit/$uuid/file/path/to/text.txt", headers = Seq(basicAuthentication), body = randomContent(22)) {
+    val headers = Seq(
+      ("Content-Disposition", "text.txt"),
```
lindareijnhoudt: the value of the Content-Disposition header is not the filename directly. 
`The first parameter [...] is always form-data. Additional parameters are case-insensitive and have arguments that use quoted-string syntax after the '=' sign. Multiple parameters are separated by a semi-colon (';').`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -128,22 +143,39 @@ class DepositServlet(app: EasyDepositApiApp) extends ProtectedServlet(app) {
...
     case e: NoSuchDepositException => NoSuchDespositResponse(e)
     case e: NoSuchFileException => NotFound(body = s"${ e.getMessage } not found")
     case e: InvalidResourceException => InvalidResourceResponse(e)
     case e: InvalidDocumentException => badDocResponse(e)
+    case e: BadRequestException => BadRequest(e.getMessage)
     case _ => internalErrorResponse(t)
   }
 
-  private def NoSuchDespositResponse(e: NoSuchDepositException) = {
+  private def NoSuchDespositResponse(e: NoSuchDepositException): ActionResult = {
```
lindareijnhoudt: Typo in NoSuchDe[s]positResponse

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -128,22 +143,39 @@ class DepositServlet(app: EasyDepositApiApp) extends ProtectedServlet(app) {
...
     // we log but don't expose which file was not found
     logger.info(e.getMessage)
```
lindareijnhoudt: should we also log who requested it?

src/test/scala/nl.knaw.dans.easy.deposit/servlets/IntegrationSpec.scala
lindareijnhoudt: From the api.yml:
`
    ContentDisposition:
      name: "Content-Disposition; filename"
      in: header
      description: the name to use for the new file in the deposit; required unless Content-Type is `application/zip`.
      required: true
      schema:
        type: "string"
`

in this spec, the equal-sign (=) seems to be missing

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -18,148 +18,217 @@ package nl.knaw.dans.easy.deposit.servlets
...
+    } yield Ok(body = toJson(depositFiles))
+      ).getOrRecoverResponse(respond)
   }
 
-  private def forDeposit[T](callback: (String, UUID) => Try[T]
-                           ): Try[T] = {
-    for {
+  post("/:uuid/file/*") { //file
+    val zip = "application/zip"
+    val bin = "application/octet-stream"
```
jo-pol: Content-type is more complex: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.17

PR #64, opened 2018-08-24
EASY-1661: Return 204 return code for successful login and logout calls

vesaakerman:
Fixes EASY-1661

#### When applied
* return code for a successful login is 204 
* return code for a successful logout is 204 

Return code when deleting with invalid bag-id should be 400, but is now 404; Return code with an invalid bag-id comes from `DepositServlet`, method `getUUID`.  Simply changing the return code for this case may result in erroneous error-code in other cases => needs to be investigated

### DIFF
--- a/src/test/scala/nl.knaw.dans.easy.deposit/authentication/AuthTestServlet.scala
+++ b/src/test/scala/nl.knaw.dans.easy.deposit/authentication/AuthTestServlet.scala
@@ -16,18 +16,19 @@
 package nl.knaw.dans.easy.deposit.authentication
 
 import nl.knaw.dans.easy.deposit.authentication.ServletEnhancedLogging._
-import org.scalatra.Ok
+import org.scalatra.NoContent
 
 class AuthTestServlet(authProvider: AuthenticationProvider) extends AbstractTestServlet(authProvider) {
 
+  // TODO: change the code so that the methods in AuthServlet are called instead of duplicating them here
   post("/login") {
     login()
-    Ok(s"signed in").logResponse
+    NoContent().logResponse
   }
 
-  put("/logout") {
+  post("/logout") {
     logOut() // destroys the scentry cookie
-    Ok("you are signed out").logResponse
+    NoContent().logResponse
   }
 }
 
--- a/src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
+++ b/src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
@@ -49,8 +49,7 @@ class SessionSpec extends TestSupportFixture with ServletFixture with ScalatraSu
       body should startWith("AuthUser(foo,List(),ACTIVE) ")
       body should endWith(" EASY Deposit API Service running")
       header("REMOTE_USER") shouldBe "foo"
-      header("Set-Cookie") should startWith("scentry.auth.default.user=")
-      header("Set-Cookie") shouldNot startWith("scentry.auth.default.user=;") // note the empty value
+      header("Set-Cookie") should startWith regex "scentry.auth.default.user=[^;].+;"
       status shouldBe OK_200
     }
   }
--- a/src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
+++ b/src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
@@ -129,13 +129,14 @@ class SessionSpec extends TestSupportFixture with ServletFixture with ScalatraSu
       uri = "/auth/login",
       params = Seq(("login", "foo"), ("password", "bar"))
     ) {
-      status shouldBe OK_200
-      body shouldBe "signed in"
-      header("Content-Type") shouldBe "text/plain;charset=UTF-8"
+      status shouldBe NO_CONTENT_204
+      body shouldBe ""
+      header("Content-Type") shouldBe "text/html;charset=UTF-8"
       header("Expires") shouldBe "Thu, 01 Jan 1970 00:00:00 GMT" // page cache
       header("REMOTE_USER") shouldBe "foo"
       val newCookie = header("Set-Cookie")
       newCookie should startWith("scentry.auth.default.user=")
+      newCookie shouldNot startWith("scentry.auth.default.user=;")
       newCookie should include(";Path=/")
       newCookie should include(";HttpOnly")
       cookieAge(newCookie) should be < 1000L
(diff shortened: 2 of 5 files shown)


rvanheest:
@DANS-KNAW/easy for review

src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
```diff
@@ -148,29 +149,30 @@ class SessionSpec extends TestSupportFixture with ServletFixture with ScalatraSu
...
       newCookie should startWith("scentry.auth.default.user=")
+      newCookie shouldNot startWith("scentry.auth.default.user=;")
       newCookie should include(";Path=/")
       newCookie should include(";HttpOnly")
       cookieAge(newCookie) should be < 1000L
     }
   }
 
-  "put /auth/logout" should "clear the cookie" in {
+  "post /auth/logout" should "clear the cookie" in {
```
rvanheest: mmm, is this testing the `AuthServlet` or the `AuthTestServlet`? I would expect the former...

src/test/scala/nl.knaw.dans.easy.deposit/servlets/HappyRoutesSpec.scala
```diff
@@ -50,9 +50,10 @@ class HappyRoutesSpec extends TestSupportFixture with ServletFixture with Scalat
       uri = "/auth/login",
       params = Seq(("login", "foo"), ("password", "bar"))
     ) {
-      status shouldBe OK_200
-      body shouldBe "signed in"
+      status shouldBe NO_CONTENT_204
+      body shouldBe ""
       header("Set-Cookie") should startWith("scentry.auth.default.user=") // details in TypicalSessionSpec
+      header("Set-Cookie") shouldNot startWith("scentry.auth.default.user=;")
```
rvanheest: what's the difference here with the line above? just the `;`? can you explain when THIS line would be the case? (in other words, when would I get `header("Set-Cookie") should startWith("scentry.auth.default.user=;")`)

src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
vesaakerman: this is testing `AuthTestServlet`

jo-pol: The question is exactly the point of the TODO above. When I initially wrote it, I did not fight long enough to find a solution. It is beyond the scope of the issue, therefore I recommended a TODO.

src/test/scala/nl.knaw.dans.easy.deposit/authentication/AuthTestServlet.scala
```diff
@@ -16,18 +16,19 @@
 package nl.knaw.dans.easy.deposit.authentication
 
 import nl.knaw.dans.easy.deposit.authentication.ServletEnhancedLogging._
-import org.scalatra.Ok
+import org.scalatra.NoContent
 
 class AuthTestServlet(authProvider: AuthenticationProvider) extends AbstractTestServlet(authProvider) {
 
+  // TODO: change the code so that the methods in AuthServlet are called instead of duplicating them here
```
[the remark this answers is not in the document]
janvanmansum: @vesaakerman : how do you propose to do that?

src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
```diff
@@ -129,13 +129,14 @@ class SessionSpec extends TestSupportFixture with ServletFixture with ScalatraSu
       uri = "/auth/login",
       params = Seq(("login", "foo"), ("password", "bar"))
     ) {
-      status shouldBe OK_200
-      body shouldBe "signed in"
-      header("Content-Type") shouldBe "text/plain;charset=UTF-8"
+      status shouldBe NO_CONTENT_204
+      body shouldBe ""
+      header("Content-Type") shouldBe "text/html;charset=UTF-8"
```
janvanmansum: Why is the content type changed to `text/html`?

src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
```diff
@@ -129,13 +129,14 @@ class SessionSpec extends TestSupportFixture with ServletFixture with ScalatraSu
...
       header("Expires") shouldBe "Thu, 01 Jan 1970 00:00:00 GMT" // page cache
       header("REMOTE_USER") shouldBe "foo"
       val newCookie = header("Set-Cookie")
       newCookie should startWith("scentry.auth.default.user=")
+      newCookie shouldNot startWith("scentry.auth.default.user=;")
```
janvanmansum: @vesaakerman : this could be more self-documenting if you would use a regular expression. What you want to say, I guess, is that there must be a non-empty scentry-cookie.

src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
janvanmansum: @jo-pol : can you comment on why the separate `TestServlet`-classes are necessary?

jo-pol: Any servlet can prove a log-in succeeded. With a TestServlet we are looser coupled.

The story behind TestAuthServlet is more complex. The constructor of AuthServlet requires EasyDepositApiApp for which we should provide a mocked AuthenticationProvider. This test was written as a proof of concept but wasn't changed into something that tested the actual class.

A quick attempt to fix it (shown below) failed. Let us keep it out of this PR.

```
  addServlet(new AuthServlet(new EasyDepositApiApp(minimalAppConfig)) {
    override def getAuthenticationProvider: AuthenticationProvider = mockedAuthenticationProvider
  }, path = "/auth/*")
```

src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
vesaakerman: that is correct.  I changed it now to a regular expression.

src/test/scala/nl.knaw.dans.easy.deposit/authentication/SessionSpec.scala
vesaakerman: The content type changed when I changed the method from `put` to `post`. The whole check is actually not needed now when the body should be empty, so I deleted the whole line.

PR #65, opened 2018-08-28
EASY-1554 unzip POSTed files

jo-pol:
Fixes EASY-1554 unzip POSTed files

#### When applied it will
* unzip files when posted as sub-request of `Content-Type multipart/form-data`
* [x] have unit tests
* upload in a staging directory, only move to draft if
  * [x] no nested zip files are detected
  * [x] no zips between plain files
* [x] merge conflicts with #67 
* [ ] replace `addPayloadFile` with  `addStagedPayloadFile` when bag-lib is released with new method https://github.com/DANS-KNAW/easy-deposit-api/blob/02ea1237df1b82e79cd1301d8565d035de1c4f99/src/main/scala/nl.knaw.dans.easy.deposit/StagedFilesTarget.scala#L49

See `readme.md` and `upload.html` in `src/test/resources/manual-test`

Try different formats (`.zip`, `.tar.gz`, a plain file with zip extension)  and character encodings. Examine the logging. Try a zip with and without a zip in it. Try to upload a zip with and without other files, if with: the zip in the first respectively in the second field.

repo        /               PR                | note
------------------------------------------- | ----
https://github.com/DANS-KNAW/dans-bag-lib/pull/42    | Should provide a method to move a payload file into place from a staging area. Perhaps a method called `addStagedPayloadFile`? Not plural to keep the caller responsible for logic concerning overwriting or rejecting files. Note that in theory an uploaded file might be an existing directory or an uploaded directory an existing file.


src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -157,11 +159,49 @@ class DepositServlet(app: EasyDepositApiApp)
...
-    managed(item.getInputStream)
-      .apply(app.writeDepositFile(_, user.id, uuid, fullPath))
-      .map((_, item))
+  private def uploadFileItem(uuid: UUID, path: Path, uploadItem: FileItem): Try[(Boolean, FileItem)] = {
+    val extensionIsZip = uploadItem.name.matches(".*.g?z(ip)?")
+    lazy val contentTypeIsZip = uploadItem.contentType.exists(_.matches(
+      "(application|multipart)/(x-)?g?zip(-compress(ed)?)?( .*)?"
+    ))
+    logger.debug(s"is it a ZIP? ${ uploadItem.name } : $extensionIsZip; ${ uploadItem.contentType } : $contentTypeIsZip ")
+    if (extensionIsZip || contentTypeIsZip)
```
rvanheest: I see a lot of common stuff here. Would this be a nice alternative?

```scala
managed(uploadItem.getInputStream)
    .apply {
        case is if extensionIsZip || contentTypeIsZip => uploadZip(uuid, path, uploadItem, is)
        case is => app.writeDepositFile(is, user.id, uuid, fullPath)
    }
    .map(_ -> uploadItem)
```

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -157,11 +159,49 @@ class DepositServlet(app: EasyDepositApiApp)
...
-    val fullPath = path.resolve(item.name)
-    managed(item.getInputStream)
-      .apply(app.writeDepositFile(_, user.id, uuid, fullPath))
-      .map((_, item))
+  private def uploadFileItem(uuid: UUID, path: Path, uploadItem: FileItem): Try[(Boolean, FileItem)] = {
+    val extensionIsZip = uploadItem.name.matches(".*.g?z(ip)?")
+    lazy val contentTypeIsZip = uploadItem.contentType.exists(_.matches(
+      "(application|multipart)/(x-)?g?zip(-compress(ed)?)?( .*)?"
+    ))
+    logger.debug(s"is it a ZIP? ${ uploadItem.name } : $extensionIsZip; ${ uploadItem.contentType } : $contentTypeIsZip ")
```
rvanheest: a question in the debug log? that doesn't really make much sense, if you ask me.
Can't we have something here like `"it is a ZIP : <etc.>"` or `"it is a XXX : <etc.>"`, depending on the specific upload.

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -157,11 +159,49 @@ class DepositServlet(app: EasyDepositApiApp)
...
+    def handleZipEntry(zipEntry: ZipEntry) = {
+      if (zipEntry.isDirectory) Success(false)
+      else {
+        logger.info(s"extracting ${ zipEntry.getName } size=${ zipEntry.getSize } crc=${ zipEntry.getCrc } method=${ zipEntry.getMethod }")
+        val fullPath = path.resolve(zipEntry.getName)
+        app.writeDepositFile(zipInputStream, user.id, uuid, fullPath)
+      }
+    }
+
+    while (
```
rvanheest: this is kind of a weird construction to see. Usually you have some condition in the while loop and then in the body, you do some things with it. However, in your case, you do everything in de 'condition' part and nothing in the body part.

This makes me wonder whether a Stream generator wouldn't suit you better here. Something like this (_not tested it, maybe you need to modify some things here or there_):

```scala
Try {
    Stream.continually(Option(zipInputStream.getNextEntry))
        .takeWhile(_.nonEmpty)
        .collect { case Some(entry) => entry } // we know now that all elements are Some(...), so we can unpack
        .map(handleZipEntry)
        .foreach {
            case Failure(e) if e.isInstanceOf[ZipException] => throw BadRequestException(s"ZIP file is malformed. ${ uploadItem.name } $e")
            case Failure(e) => throw e
            case Success(_) => // do nothing
        }
}.map(_ => false)
```

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -157,11 +159,49 @@ class DepositServlet(app: EasyDepositApiApp)
...
+      Try(zipInputStream.getNextEntry)
+        .map(Option(_).map(handleZipEntry)) match {
+        case Success(None) => false // end of zip
+        case Success(Some(Success(_))) => true // extracted and uploaded
+        case Success(Some(Failure(e))) => return Failure(e) // could not save
+        case Failure(e) if e.isInstanceOf[ZipException] => // could not extract TODO still fail fast? Other files might hav been uploaded.
+          return Failure(BadRequestException(s"ZIP file is malformed. ${ uploadItem.name } $e"))
+        case Failure(e) => return Failure(e)
+      }) {}
+    Success(false)
```
rvanheest: do you always return `false`??? Why is that? It doesn't make much sense to me. If you just want to communicate `Success` or `Failure`, why not use `Try[Unit]` here?

src/test/scala/nl.knaw.dans.easy.deposit/servlets/HappyRoutesSpec.scala
```diff
@@ -169,7 +169,7 @@ class HappyRoutesSpec extends TestSupportFixture with ServletFixture with Scalat
       headers = Seq(("Authorization", fooBarBasicAuthHeader))
     ) {
       status shouldBe BAD_REQUEST_400
-      body shouldBe """Bad Request. invalid DatasetMetadata: class java.lang.Exception don't recognize {"title":"blabla"}"""
+      body shouldBe """Bad Request. invalid DatasetMetadata: java.lang.Exception: don't recognize {"title":"blabla"}"""
```
rvanheest: is this a message that goes to the user/client? I would much rather not have exception class names like `java.lang.Exception` in there. Just a **very clear** error message.

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -129,13 +128,12 @@ class DepositServlet(app: EasyDepositApiApp)
     (for {
       uuid <- getUUID
       path <- getPath
-      _ <- isMultipart
-      newFileItems <- getFileItems
+      _ <- getContentTypeIfMultipart
+      newFileItems <- fileMultiParams.values.flatten // TODO fileParams.values gives the first file of each form-field, why/when/which
```
rvanheest: is `.valuesIterator.flatten` not lazy enough? In that case you don't need the `toStream` here.
Not sure though, haven't tested it...

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -159,49 +157,28 @@ class DepositServlet(app: EasyDepositApiApp)
       ).getOrRecoverResponse(respond)
   }
 
-  private def uploadFileItem(uuid: UUID, path: Path, uploadItem: FileItem): Try[(Boolean, FileItem)] = {
+  private def uploadFileItem(uuid: UUID, path: Path, uploadItem: FileItem): Try[FileItem] = {
+    managed(uploadItem.getInputStream).apply[Try[_]] {
```
rvanheest: The types don't really line up, do they? `unzipDepositFile` has `Try[Unit]` while the second case has `Try[FileItem]`. It is therefore probably not correct to say in the return type that it has type `Try[FileItem]`... Any clue how you could fix that?

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -159,49 +157,28 @@ class DepositServlet(app: EasyDepositApiApp)
...
-    def handleZipEntry(zipEntry: ZipEntry) = {
-      if (zipEntry.isDirectory) Success(false)
-      else {
-        logger.info(s"extracting ${ zipEntry.getName } size=${ zipEntry.getSize } crc=${ zipEntry.getCrc } method=${ zipEntry.getMethod }")
-        val fullPath = path.resolve(zipEntry.getName)
-        app.writeDepositFile(zipInputStream, user.id, uuid, fullPath)
-      }
+  private def getContentTypeIfMultipart = {
+    val multiPart = "multipart/"
+    request.getHeader("Content-Type").blankOption match {
```
rvanheest: would it look better if you would do this match using a `map` and [`getOrRecover`](https://github.com/DANS-KNAW/dans-scala-lib/blob/c7edb07a0331ecc3ffaacebc9a058408154278b5/src/main/scala/nl/knaw/dans/lib/error/package.scala#L160)?

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
jo-pol: Here I only need `Try[Unit]` for both cases. The result of `writeDepositFile` is required for another call:

https://github.com/DANS-KNAW/easy-deposit-api/blob/6a474d3d7683644d8890a9fc12a540840fa23c74/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala#L145

The logic is subject to change anyhow because of #66 : one zip only or none. The current logic unzips any zip.

rvanheest: Mmm, maybe I'm a bit confused here; are you saying that `writeDepositFile` is used twice (perfectly OK!), but that in this instance you only need the `Unit` return type? That isn't reflected by the return type of `uploadFileItem`. Anyway, I let you play with it. Just noticed this type conflict 😃

rvanheest:
See https://github.com/DANS-KNAW-jp/easy-deposit-api/pull/2 for the fix in https://github.com/DANS-KNAW/easy-deposit-api/blob/a42a89020464b75184822ae1a03d28d7b21e3e32/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala#L92

rvanheest:
Another test case: a zip file with a file-folder structure; not just 3 files in the root of the zip. Maybe you have that one already, but I couldn't find it yet (just looked very quickly...)

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -128,13 +129,14 @@ class DepositServlet(app: EasyDepositApiApp)
...
       _ <- isMultipart
-      newFileItems <- getFileItems
-        .toStream
-        .withFilter(_.name.blankOption.isDefined)
-        .map { fileItem => postFileItem(uuid, path, fileItem) }
-        .failFast
-        .map(_.withFilter(_._1).map(_._2))
-    } yield Ok() // TODO create multiple location headers from newFileItems?
+      fileItems = fileMultiParams.valuesIterator.flatten.buffered
+      _ = while (fileItems.head.name.trim.isEmpty) { fileItems.next() } // skip leading form fields without selected files
```
rvanheest: you're probably looking for [`fileItems.dropWhile(_.name.trim.isEmpty)`](https://www.scala-lang.org/api/current/scala/collection/Seq.html#dropWhile(p:A=%3EBoolean):Repr) and then the rest of the old code.

jo-pol: No, head of a buffered iterator is peeking without forwarding the pointer.

rvanheest: I know, but `fileItems.next()` IS forwarding the pointer

jo-pol:
@lindareijnhoudt @janvanmansum 

Please your opinions about 
* nested zips 
* hidden files as the __MACOSX  directory on https://github.com/DANS-KNAW/easy-deposit-api/blob/2a79d5a189933c57c13d3ea6102d6d782b312e28/src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala#L176

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
jo-pol: But it should only be forwarded if the condition applies, not before it is checked.

jo-pol:
@rvanheest is referring to "Comparative Exploration of Document Collections", perhaps an interesting project but off topic for this module. 

We are an archive, I remember we had the discussion before whether or not to store OS-specific hidden files/directories but don't remember the conclusion. I suppose there are also other files/folders  than __MACOSX.

jo-pol:
Once this is merged I can create the real PR for https://github.com/DANS-KNAW-jp/easy-deposit-api/pull/3

PR #66, opened 2018-08-28
NO JIRA. Amended POST API.

janvanmansum:
The previous version was a bit of a hack, as it documented a part-header at the level where for the rest http-message headers are documented. This is not necessary at all. The Swagger documentation [gives clear guidance on how to document multipart/form-data](https://swagger.io/docs/specification/describing-request-body/file-upload/).

Also, some functionality was specified that I disagree with, like unzipping ZIP-files in a multipart/form-data request and looking at the extension of a file name to determine whether a file is a ZIP. I strongly suggest we try to keep close to standards, best practices and in general using things the way they were intended.

#### When applied it will
* Amend the specs for POST-ing to directories

`api.yml`

### DIFF
--- a/docs/api/api.yml
+++ b/docs/api/api.yml
@@ -1,10 +1,10 @@
 openapi: 3.0.1
 
 servers:
-- description: Debug environment
-  url: http://localhost:20190/
-- description: Local Vagrant machine
-  url: http://test.dans.knaw.nl
+  - description: Debug environment
+    url: http://localhost:20190/
+  - description: Local Vagrant machine
+    url: http://test.dans.knaw.nl
 info:
   description: Public interface for depositing datasets in EASY.
   version: 1.0.0
@@ -109,7 +109,7 @@ paths:
         500:
           $ref: "#/components/responses/InternalServerError"
       security:
-      - BearerAuth: []
+        - BearerAuth: []
 
   /deposit:
     get:
@@ -267,7 +267,7 @@ paths:
         500:
           $ref: "#/components/responses/InternalServerError"
       security:
-      - BearerAuth: []
+        - BearerAuth: []
 
   /deposit/{id}/state:
     get:
@@ -368,12 +368,12 @@ paths:
                 items:
                   $ref: "#/components/schemas/FileInfo"
                 example:
-                - filename: myfile.txt
-                  dirpath: path/to/parent/dir
-                  sha1sum: ab3aa0555f31a8d7809fae4b03a95195edb01f5c
-                - filename: myfile2.txt
-                  dirpath: path/to/parent/dir
-                  sha1sum: ab3aa0555f31a8d7809fae4b03a95195edb01f5d
+                  - filename: myfile.txt
+                    dirpath: path/to/parent/dir
+                    sha1sum: ab3aa0555f31a8d7809fae4b03a95195edb01f5c
+                  - filename: myfile2.txt
+                    dirpath: path/to/parent/dir
+                    sha1sum: ab3aa0555f31a8d7809fae4b03a95195edb01f5d
 
         401:
           $ref: "#/components/responses/Unauthorized"
@@ -386,7 +386,7 @@ paths:
     post:
       tags:
       - directory
-      summary: Adds one or more files or a subdirectory.
+      summary: Adds one or more files.
       description: |
         Depending on the `Content-Type` header one or multiple files can be sent. The files will
         be added in the directory pointed to by `dir_path`.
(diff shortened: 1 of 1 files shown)


docs/api/api.yml
```diff
@@ -386,29 +386,32 @@ paths:
     post:
       tags:
       - directory
-      summary: Adds a file or a subdirectory.
+      summary: Adds one or more files.
```
lindareijnhoudt: so how would you add a subdirectory, without a specific file?

docs/api/api.yml
```diff
@@ -386,29 +386,32 @@ paths:
...
+        The `Content-Type` must always be `multipart/form-data`, which means the body is formatted using the
+        [multipart media type](https://tools.ietf.org/html/rfc2046.html#section-5.1); each part contains one file.
+        The header area of each part specifies the file name using a `Content-Disposition` header.
+
+        If the body contains only one part and that part has `Content-Type: application/zip`, then the body
+        is a file in ZIP format. This ZIP file can contain a directory hierarchy. The ZIP file will be
+        unzipped in `dir_path`. For each file in the ZIP file, the path in the deposit will be `dir_path`/`path_in_zip`.
+
+        In all other cases the files are all stored in the directory pointed to by `dir_path` and it is not
+        allowed to include any files with `Content-Type: application/zip`. If multiple ZIP Note, that this means that only a flat
```
lindareijnhoudt: not a sentence ...

docs/api/api.yml
```diff
@@ -386,29 +386,32 @@ paths:
...
             schema:
-              type: string
-              format: binary
+              type: array
+              items:
+                type: string
+                format: binary
       responses:
         201:
           description: file or subdirectory added
```
lindareijnhoudt: the 'adding a subdirectory' was dropped from the description. is that still possible?

docs/api/api.yml
```diff
@@ -386,29 +386,32 @@ paths:
...
       - directory
-      summary: Adds a file or a subdirectory.
+      summary: Adds one or more files.
       description: |
-        Sub request of a request with `Content-Type` value `multipart/form-data`.
-        By default the files listed in the body of the main request are saved in `dir_path` under the file name provided
-        in the `Content-Disposition` header. However, if a file has the extension ZIP, it
-        is extracted in `dir_path` and `Content-Disposition` is ignored (and may be left out).
-        If `dir_path`, or any of its parents do not exist yet, they are first created.
+        The `Content-Type` must always be `multipart/form-data`, which means the body is formatted using the
```
jo-pol: What type of response code do you want for other content types? #63 is causing 'not-implemented'

janvanmansum: The proper response is 406 Not acceptable. "The requested resource is capable of generating only content not acceptable according to the Accept headers sent in the request." See [the wikipedia page on http status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors).

docs/api/api.yml
janvanmansum: There is no way to just create a subdirectory, except by sending a ZIP file that contains one empty directory.

docs/api/api.yml
janvanmansum: I think 400 Bad request is the best option here.

docs/api/api.yml
janvanmansum: 👍

jo-pol (approved):

docs/api/api.yml
```diff
@@ -386,38 +386,45 @@ paths:
...
+        [multipart media type](https://tools.ietf.org/html/rfc2046.html#section-5.1); each part contains one file.
+        The header area of each part specifies the file name using a `Content-Disposition` header.
+
+        If the body contains only one part and that part has `Content-Type: application/zip`, then the body
+        is a file in ZIP format. This ZIP file can contain a directory hierarchy. The ZIP file will be
+        unzipped in `dir_path`. For each file in the ZIP file, the path in the deposit will be `dir_path`/`path_in_zip`.
```
jo-pol: What to do if the n-th file appears to be a ZIP? As the list of files is provided via an iterator, it is not logical to examine them all before processing them.

docs/api/api.yml
jo-pol: Seeing https://github.com/scalatra/scalatra/blob/2a038af00bea06e1bc97e328be92ad9ec7ccf3b5/core/src/main/scala/org/scalatra/servlet/FileUploadSupport.scala#L101

I guess this should be: `Content-Type` must always start with `multipart/`

jo-pol (changes requested):

docs/api/api.yml
janvanmansum: I think is must be `multipart/form-data` specifically. There are other types of `multipart`, see https://tools.ietf.org/html/rfc2046.html#section-5.1. I am not sure if we can or need to handle those as well. I specified it so that only `multipart/form-data` is supported, but if you find something in the document I mentioned above that makes you think we should support more than that, let me know.

docs/api/api.yml
janvanmansum: The way it is specified now, the operation must then fail.

lindareijnhoudt: should the files that were uploaded so far be deleted, or should all files be uploaded, except for the erroneous zipfile?

janvanmansum: So no second thought then ;)

PR #67, opened 2018-09-05
EASY-1681: Return single FileInfo object for GET on specific file

vesaakerman:
Fixes EASY-1681

#### When applied it will
* return information about a file when the path in GET /deposit/{id}/file/{path} points to a file

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
@@ -34,6 +34,28 @@ import scala.util.{ Failure, Success, Try }
  */
 case class DataFiles(bag: DansBag) extends DebugEnhancedLogging {
 
+  val dataDir = bag.baseDir / "data"
+
+  /**
+   * Returns 'true' if the path points to a directory.
+   *
+   * @param path a relative path.
+   * @return 'true' if directory, else 'false'
+   */
+  def isDirectory(path: Path): Boolean = {
+    dataDir / path.toString isDirectory
+  }
+
+  /**
+   * Returns contents of a file.
+   *
+   * @param path a relative path to a data file.
+   * @return contents of the file
+   */
+  def fileContents(path: Path): Try[String] = Try {
+    dataDir / path.toString contentAsString
+  }
+
   /**
    * Lists information about the files the directory `path` and its subdirectories.
    * A previous crash during a recursive delete process (deleting a directory),
@@ -44,7 +66,7 @@ case class DataFiles(bag: DansBag) extends DebugEnhancedLogging {
    * @return a list of [[FileInfo]] objects
    */
   def list(path: Path = Paths.get("")): Try[Seq[FileInfo]] = {
-    val parentPath = bag.baseDir / "data" / path.toString
+    val parentPath = dataDir / path.toString
     val manifestMap = bag.payloadManifests
 
     def convert(items: Map[File, String]) = {
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
@@ -92,10 +92,11 @@ case class DataFiles(bag: DansBag) extends DebugEnhancedLogging {
    */
   def fileInfo(path: Path = Paths.get("")): Try[FileInfo] = {
     val manifestMap = bag.payloadManifests
+    val manifests = manifestMap.get(SHA1).orElse(manifestMap.values.headOption)
     val absolutePath = dataDir / path.toString
-    val fileExists = manifestMap.get(SHA1).orElse(manifestMap.values.headOption).get.contains(absolutePath)
+    val fileExists = manifests.get.contains(absolutePath)
     if (fileExists) {
-      val checksum = manifestMap.get(SHA1).orElse(manifestMap.values.headOption) get absolutePath
+      val checksum = manifests get absolutePath
       Success(FileInfo(path.getFileName.toString, bag.data.relativize(absolutePath), checksum))
     }
     else {
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
@@ -25,6 +25,7 @@ import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 
 import scala.util.{ Failure, Success, Try }
 import scala.language.postfixOps
+
 /**
  * Represents the data files of a deposit. The data files are the content files that the user uploads,
  * i.e. the files that are the actual target of preservation. The dataset metadata is ''not'' included
@@ -81,6 +82,25 @@ case class DataFiles(bag: DansBag) extends DebugEnhancedLogging {
       .orElse(manifestMap.values.headOption)
       .map(items => Success(convert(items)))
       .getOrElse(Failure(new Exception(s"no algorithm for ${ bag.baseDir }")))
+  }
+
+  /**
+   * Lists information about a file.
+   *
+   * @param path a relative path into a data files.
+   * @return a [[FileInfo]] object
+   */
+  def fileInfo(path: Path = Paths.get("")): Try[FileInfo] = {
+    val manifestMap = bag.payloadManifests
+    val absolutePath = dataDir / path.toString
+    val fileExists = manifestMap.get(SHA1).orElse(manifestMap.values.headOption).get.contains(absolutePath)
+    if (fileExists) {
+      val checksum = manifestMap.get(SHA1).orElse(manifestMap.values.headOption) get absolutePath
+      Success(FileInfo(path.getFileName.toString, bag.data.relativize(dataDir / path.toString), checksum))
+    }
+    else {
+      Failure((new NoSuchFileException(s"file ${path.getFileName}")))
+    }
   }
 
   private def toFileInfo(file: File, checksum: String): FileInfo = {
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
@@ -23,8 +23,8 @@ import nl.knaw.dans.bag.ChecksumAlgorithm.SHA1
 import nl.knaw.dans.bag.DansBag
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 
-import scala.util.{ Failure, Success, Try }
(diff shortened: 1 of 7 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
```diff
@@ -61,6 +84,25 @@ case class DataFiles(bag: DansBag) extends DebugEnhancedLogging {
...
+   *
+   * @param path a relative path into a data files.
+   * @return a [[FileInfo]] object
+   */
+  def fileInfo(path: Path = Paths.get("")): Try[FileInfo] = {
+    val manifestMap = bag.payloadManifests
+    val absolutePath = dataDir / path.toString
+    val fileExists = manifestMap.get(SHA1).orElse(manifestMap.values.headOption).get.contains(absolutePath)
+    if (fileExists) {
+      val checksum = manifestMap.get(SHA1).orElse(manifestMap.values.headOption) get absolutePath
```
janvanmansum: Questions:
* What will happen if there is no payload manifest at all?
* What are the invariants that we can count on?
* What do you report if an invariant is violated? 
* Would it make sense to check the file system to see if the file actually exists? What would you report if the file then exists on disk, but there is no manifest entry for it?

Comments:
* Lines 96 and 98 duplicate quite a lot.
* I see a lot of overlap with `fileInfoSeq`.

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -191,22 +195,21 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
    */
   def writeDataMetadataToDeposit(dm: DatasetMetadata, user: String, id: UUID): Try[Unit] = for {
-    deposit <- DepositDir.get(draftsDir, user, id)
+    deposit <- getDeposit(user, id)
     _ <- deposit.writeDatasetMetadataJson(dm)
   } yield ()
 
   /**
-   * Returns the list of [[FileInfo]] objects for the dataset files in this deposit.
+   * Returns dataset files of the deposit
```
janvanmansum: Why did you change the description?

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -119,8 +118,10 @@ class DepositServlet(app: EasyDepositApiApp)
     (for {
       uuid <- getUUID
       path <- getPath
-      depositFiles <- app.getDepositFiles(user.id, uuid, path)
-    } yield Ok(body = toJson(depositFiles))
+      dataFiles <- app.getDataFiles(user.id, uuid)
```
janvanmansum: This type of logic (deciding between directory? or single file?) should really be handled in the application object.

src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
```diff
@@ -43,8 +66,8 @@ case class DataFiles(bag: DansBag) extends DebugEnhancedLogging {
    * @param path a relative path into data files directory of the bag.
    * @return a list of [[FileInfo]] objects
    */
-  def list(path: Path = Paths.get("")): Try[Seq[FileInfo]] = {
-    val parentPath = bag.baseDir / "data" / path.toString
+  def fileInfoSeq(path: Path = Paths.get("")): Try[Seq[FileInfo]] = {
```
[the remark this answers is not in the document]
vesaakerman: By a request from @jo-pol , to show that the two methods (fileInfo and fileInfoSeq) are closely related.  I agree with Joke.

src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
```diff
@@ -34,6 +35,28 @@ import scala.util.{ Failure, Success, Try }
...
+   * Returns 'true' if the path points to a directory.
+   *
+   * @param path a relative path.
+   * @return 'true' if directory, else 'false'
+   */
+  def isDirectory(path: Path): Boolean = {
+    dataDir / path.toString isDirectory
+  }
+
+  /**
```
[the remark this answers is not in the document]
vesaakerman: It is used...

src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
```diff
@@ -34,6 +35,28 @@ import scala.util.{ Failure, Success, Try }
  */
 case class DataFiles(bag: DansBag) extends DebugEnhancedLogging {
 
+  val dataDir = bag.baseDir / "data"
```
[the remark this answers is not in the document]
vesaakerman: Right, I've made it private now.

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
vesaakerman: Because the method didn't return anymore [FileInfo] objects, but deposit data files.

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
vesaakerman: I have now moved the distinction between directories and files into `EasyDepositApiApp`

src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
vesaakerman: In my understanding the bag is already validated. It is possible still to check whether `manifestMap` is `null`, and if it is so, to throw an exception, and take this exception in account in `DepositServlet`.  But I am wondering if that is a bit overkill, if the deposit indeed has already been validated?  The same applies to the question whether we would check if the  file exists in the filesystem, even if it is absent in the manifest file.

vesaakerman: The lines 96 and 98 were indeed very similar, thus I modified the code a bit

jo-pol:
I now see @janvanmansum asked for this change. https://github.com/DANS-KNAW/easy-deposit-api/pull/67#discussion_r216232451

The problem I tried to point out, according to the swagger spec the response body should be constructed with a `Seq[FileInfo].toJson` in case of a directory and a `FileInfo.toJson` in case of a file. So either the spec should change or we do need to call different methods from the app class. 

@lindareijnhoudt I don't know whether or how this is already implemented in the UI.

jo-pol:
Pleas next time trim your post when replying via email, or adjust the recipient.
I now see you are returning and object, did not expect that.

src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
```diff
@@ -34,6 +35,28 @@ import scala.util.{ Failure, Success, Try }
  */
 case class DataFiles(bag: DansBag) extends DebugEnhancedLogging {
 
+  private val dataDir = bag.baseDir / "data"
```
jo-pol: This is the same as `bag.data`. I prefer the latter. so drop this one.

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -119,8 +118,9 @@ class DepositServlet(app: EasyDepositApiApp)
...
+      contents <- if (dataFiles.isDirectory(path)) dataFiles.list(path) else dataFiles.fileInfo(path)
```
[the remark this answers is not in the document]
janvanmansum: @vesaakerman @jo-pol : I'd personally prefer just `list` and `get`. This short style was there from the beginning. Methods should normally have a name that is or starts with a verb.

src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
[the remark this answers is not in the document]
janvanmansum: I found the comment with @jo-pol 's request. I disagree. I think we should try to keep the style uniform and also adhere to standard conventions in naming methods.

src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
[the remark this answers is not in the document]
janvanmansum: I am referring to `fileContents` (which should by the way be called `getFileContents`). Where is it used, my IntelliJ can't locate the caller.

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
janvanmansum: @vesaakerman @jo-pol OK, that looks better. `getDataFiles` can now be made private by the way. It is however a pity that the return type of getFileInfo is now `Try[Object]`.  Here it *would* make sense to have sepate `listFileInfos` and `getFileInfo` methods.

src/main/scala/nl.knaw.dans.easy.deposit/DataFiles.scala
janvanmansum: @vesaakerman @jo-pol It is not a big deal to me, but there seems little value in this field. Any arguments in favor or contra (apart from preferences)?

jo-pol: The field used to be valuable before the new bag-lib was used. Now that we also have `bag.data` it is confusing to use both. Only by this review I saw I should have cleaned it up when introducing the bag-lib.

src/test/scala/nl.knaw.dans.easy.deposit/servlets/IntegrationSpec.scala
```diff
@@ -117,6 +115,31 @@ class IntegrationSpec extends TestSupportFixture with ServletFixture with Scalat
...
+
+    // get the file information
+    expectsUserFooBar
+    get(uri = s"/deposit/$uuid/file/path", headers = Seq(basicAuthentication)) {
+      status shouldBe OK_200
+      body shouldBe """[{"fileName":"text.txt","dirPath":"path/to","sha1sum":"c5b8de8cc3587aef4e118a481115391033621e06"}]"""
+    }
+  }
+
+  "scenario: POST /deposit; GET /deposit/:uuid/file/path/to/text.txt" should "return information of the file" in {
```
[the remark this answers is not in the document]
vesaakerman: I have reworded the comments;  I put the 'twice' in the right place, and also changed the text what the test should return.

PR #68, opened 2018-09-27
NO JIRA issue: DOI bug, identifiers in DDM

jo-pol:
### DOI bug

The UI does not see that a failing submit may have assigned a DOI. Thus the next submit attempt fails because the DOI gets erased in the `metadata/dataset.json` while still present in the `deposit.properties`.

### Solution

Rather than going through complex error handling and roll-back problems, the UI should go through the following steps when a user pushes the submit button:
* GET the DOI if not done by the user before
* PUT the metadata with the generated DOI
* PUT the new state
---
#### When applied
* The api will no longer assign a DOI on submit. In stead the api will fail if it the DOI missing.
* Copy the identifiers from json to DDM
* Give more json centric error messages on missing fields

See `src/test/resources/manual-test`

### DIFF
--- a/docs/api/api.yml
+++ b/docs/api/api.yml
@@ -300,6 +300,8 @@ paths:
       description: |
         The client is only allowed to update the deposit from DRAFT to SUBMITTED or
         from REJECTED to DRAFT. Other state transitions will be effected by back-end processes.
+        The client must have called `GET /deposit/{id}/doi` and `PUT /deposit/{id}/metadata`
+        with all fields that are madatory at submission.
       parameters:
       - $ref: "#/components/parameters/DepositId"
       operationId: updateDepositState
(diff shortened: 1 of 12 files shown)


docs/api/api.yml
```diff
@@ -300,6 +300,8 @@ paths:
       description: |
         The client is only allowed to update the deposit from DRAFT to SUBMITTED or
         from REJECTED to DRAFT. Other state transitions will be effected by back-end processes.
+        The client must have called `GET /deposit/{id}/doi` and `PUT /deposit/{id}/metadata`
```
lindareijnhoudt: something like this?

In order to submit, the deposit needs to be valid. In particular, the json metadata MUST be valid, with all mandatory fields specified, among which the Digital Object Identifier (DOI).

PR #69, opened 2018-10-05
EASY-1735 + EASY-1736 DDM convert json to XML including subject/coverages

jo-pol:
Fixes EASY-1735

#### When applied it will
* convert most json fields to dataset.xml at submission, ~~still to do:~~
  * [x] subject en coverages  https://drivenbydata.atlassian.net/browse/EASY-1736
  * [x] author IDs
* coordinates for spatial boxes/point changed from number to strings with number format
* [x] exhaustive tests for edge cases
* [x] informative error messages on input errors

Submit a dataset with all kinds and variations of content for the metadata, see also `src/test/resources/manual-test`


src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
```diff
@@ -15,52 +15,50 @@
...
-                           subjects: Option[Seq[PossiblySchemedKeyValue[String]]] = None,
-                           alternativeIdentifiers: Option[Seq[SchemedValue[String]]] = None,
+                           audiences: Option[Seq[SchemedKeyValue]] = None,
+                           subjects: Option[Seq[PossiblySchemedKeyValue]] = None, //TODO xml
+                           alternativeIdentifiers: Option[Seq[SchemedValue]] = None,
                            relations: Option[Seq[RelationType]] = None,
-                           languagesOfFiles: Option[Seq[PossiblySchemedKeyValue[String]]] = None,
-                           dates: Option[Seq[QualifiedSchemedValue[String, DateQualifier]]] = None,
+                           languagesOfFiles: Option[Seq[PossiblySchemedKeyValue]] = None,
+                           dates: Option[Seq[Date]] = None,
```
rvanheest: make this and other fields that require further analysis private

jo-pol: causes problems for testing:
https://github.com/DANS-KNAW/easy-deposit-api/blob/3e6b9fb3ea8344c914517b31a49d14daaa2825fc/src/test/scala/nl.knaw.dans.easy.deposit/docs/DDMSpec.scala#L184

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadataSpec.scala
```diff
@@ -148,66 +159,100 @@ class DatasetMetadataSpec extends TestSupportFixture {
...
+  }
+
+  it should "reject an audience with an empty string for a value" in {
+    """{"audiences": [{ "scheme": "xxx", "key": "yyy", "value": "" }]}"""
+      .causesInvalidDocumentException("""requirement failed: Empty string for a mandatory field; got {"scheme":"xxx","key":"yyy","value":""} SchemedKeyValue""")
+  }
+
+  "DatasetMetadata.spatialBoxes" should "reject a non-numeric value" in {
+    """{"spatialBoxes": [ { "scheme": "RD", "north": "486890,5", "east": 121811.88, "south": 436172.5,  "west": 91232.016 }]}""".stripMargin
+      .causesInvalidDocumentException("""requirement failed: Invalid number [486890,5]; got {"scheme":"RD","north":"486890,5","east":"121811.88","south":"436172.5","west":"91232.016"} SpatialBox""")
```
jo-pol: @rvanheest @lindareijnhoudt 
Note that the UI can but does not need to quote the numbers!

jo-pol: But it will get them quoted from the server.

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
```diff
@@ -116,14 +116,21 @@ object DatasetMetadata {
     }
   }
 
-  trait RequiresNonEmpty {
-    def requireNonEmptyString[T](value: T, tag: String): Unit = {
+  trait Requirements {
```
rvanheest: Why not extract this trait into its own class. Especially since you use it in so many other classes.

src/main/scala/nl.knaw.dans.easy.deposit/docs/dm/Author.scala
```diff
@@ -46,7 +44,8 @@ case class Author(titles: Option[String] = None,
       case (true, true) => s"$name; ${ organization.getOrElse("") }"
```
rvanheest: if you would match on the `Some`/`None` instances, rather than on `Boolean`s, this piece of code would:
* be much more readable
* not require `getOrElse`

rvanheest: Regarding the `.filter(...).map(...)` above: could you fit this into a `collect`? Looks like this is a great use case for it and it reduces the number of iterations through the list.

src/main/scala/nl.knaw.dans.easy.deposit/package.scala
```diff
@@ -70,7 +70,7 @@ package object deposit {
     }
   }
 
-  implicit class StringExtensions(val s: String) {
+  implicit class StringExtensions(val s: String) extends AnyVal {
```
rvanheest: Could you add a `// TODO` to this implicit class, saying that this is no longer needed once we install better-files v3.6 ([link 1](https://github.com/pathikrit/better-files/blob/master/CHANGES.md#v360), [link 2](https://github.com/pathikrit/better-files/commit/5afb5f1ac58b248582e5cffcd8f32ebb2d91cd83)).

I'll probably get to that upgrade early next week, so stay tuned!

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadataSpec.scala
rvanheest: @jo-pol thanks for the heads up!

> Note that the UI can but does not need to quote the numbers!

As of https://github.com/DANS-KNAW/easy-deposit-ui/pull/84 we changed this to align with this PR: it will serve strings to the client, rather than numbers

> But it will get them quoted from the server.

[That was already in place from the beginning.](https://github.com/DANS-KNAW/easy-deposit-ui/blob/c37d0ffedc26c1ea05d4693c68d95b9b743462ca/src/main/typescript/lib/metadata/SpatialBox.ts#L36-L39)

src/main/scala/nl.knaw.dans.easy.deposit/docs/dm/Author.scala
```diff
@@ -0,0 +1,50 @@
...
+  require(!hasRedundant, buildMsg(redundantMsg))
+
+  def isRightsHolder: Boolean = role.exists(_.key == "RightsHolder")
+
+  override def toString: String = { // TODO ID's for rightsHolders
+    def name = Seq(titles, initials, insertions, surname)
+      .collect{case Some(s) if s.trim.nonEmpty => s}
+      .mkString(" ")
+
+    (surname.find(_.trim.nonEmpty), organization.find(_.trim.nonEmpty)) match {
```
jo-pol: @rvanheest with your way I end up with two ways of treating a `Some("   \t\n")`
* `.isProvided`
* filter/collect `.trim.nonEmpty`

jo-pol: And I'm also using the approach below to turn Some's with empty strings into None's. This approach can become elaborate and easy to miss when a field is added some day.
https://github.com/DANS-KNAW/easy-deposit-api/blob/ec1c8a08c2eb0d3183e811a18ca8fd780e394437/src/main/scala/nl.knaw.dans.easy.deposit/docs/dm/Relation.scala#L55-L58

src/test/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadataSpec.scala
```diff
@@ -129,85 +140,142 @@ class DatasetMetadataSpec extends TestSupportFixture {
...
-    DatasetMetadata(s) shouldBe a[Success[_]]
+    DatasetMetadata("""{ "relations": [ { "qualifier": "dcterms:hasFormat", "title": "string" } ] }""") shouldBe a[Success[_]]
   }
 
   it should "reject a relation with just a qualifier" in {
-    val s: JsonInput = """{ "relations": [ { "qualifier": "dcterms:hasFormat" } ] }"""
-    shouldReturnCustomMessage(s, """expected one of (Relation | RelatedIdentifier) got: {"qualifier":"dcterms:hasFormat"}""")
+    """{ "relations": [ { "qualifier": "dcterms:hasFormat" } ] }"""
+      .causesInvalidDocumentException( """expected one of (Relation | RelatedIdentifier) got: {"qualifier":"dcterms:hasFormat"}""")
+  }
```
jo-pol: **Too big for the scope of the pull request**
Shouldn't we keep the barbarians at the gate and somehow escape strings that are vulnerable to injection? The following unit test shows just one simple case that now gets through. 

```suggestion
  }
  
    private val valueWithHtmlInjection = """<a href="javascript:alert('hello')">Foo<a> bar"""
  "values with injected html" should behave like validDatasetMetadata(
    input = Try(new MinimalDatasetMetadata(alternativeTitles = Some(Seq(valueWithHtmlInjection)))),
    subset = actualDDM => dcmiMetadata(actualDDM),
    expectedDdmContent =
      <ddm:dcmiMetadata>
        <dcterms:identifier xsi:type="id-type:DOI">mocked-DOI</dcterms:identifier>
        <dcterms:alternative>{ valueWithHtmlInjection }</dcterms:alternative>
        <dcterms:dateSubmitted xsi:type="dcterms:W3CDTF">2018-03-22</dcterms:dateSubmitted>
      </ddm:dcmiMetadata>
  )
```

PR #74, opened 2018-11-06
Easy 1781

stephanHogenboom:
Fixes EASY-1781

#### When applied it will
* add the line identifier.doi.registered=no in the deposit.properties file of new deposits

DepositDir class

create new deposit and verify that the field is added to the properties file

### DIFF
--- a/src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
+++ b/src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
@@ -140,7 +140,8 @@ class DepositDirSpec extends TestSupportFixture with MockFactory {
 
     // preconditions
     mdFile.contentAsString shouldBe "{}"
-    oldDepositProperties should not include "identifier.doi"
+    oldDepositProperties should include("identifier.doi.registered = no")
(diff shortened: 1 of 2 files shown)


src/test/scala/nl.knaw.dans.easy.deposit/DepositDirSpec.scala
```diff
@@ -140,7 +140,8 @@ class DepositDirSpec extends TestSupportFixture with MockFactory {
 
     // preconditions
     mdFile.contentAsString shouldBe "{}"
-    oldDepositProperties should not include "identifier.doi"
+    oldDepositProperties should include("identifier.doi.registered = no")
```
rvanheest: Besides the fact that I would rather see this `depositPropertiesFile` be read as a `PropertiesConfiguration` rather than a `String` and doing `props.getString(...) shouldBe ...` (@jo-pol, I'm looking at you 😉), I am wondering what this check adds to this test. You're testing that it says `identifier.doi.registered=no`, but maybe there is a better place for this, as it seems a bit out of context in this test. Maybe @jo-pol can point you to a nice place to put this line?

PR #78, opened 2018-11-19
EASY-1693 not found responses

jo-pol:
Fixes EASY-1693

#### When applied it will
* fail fast if the deposit is not found at all
* not return absolute paths in NotFound responses
* not mix up not found of  "deposit" or "something within a deposit"
* allow repeated requests to get a DOI without saving the metadata in between these requests
* have consistent error messages for DOI problems

the new test class and api.yml


src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -120,22 +120,19 @@ case class DepositDir private(baseDir: File, user: String, id: UUID) extends Deb
...
       .map(_.titles.flatMap(_.headOption).getOrElse(""))
-      .recoverWith {
-        case _: NoSuchDepositException => Success("")
-        case t => Failure(t)
-      }
   }
 
   private def getDepositProps = Try {
     new PropertiesConfiguration(depositPropertiesFile.toJava)
+  }.map { props =>
```
rvanheest: * use `flatMap`, `Success` and `Failure` for clearer code
* use a pattern match like `.flatMap { case props if props.getKeys.hasNext => Success(props); case _ => Failure(...) }`

src/test/scala/nl.knaw.dans.easy.deposit/servlets/NotFoundSpec.scala
```diff
@@ -0,0 +1,182 @@
...
+import nl.knaw.dans.easy.deposit.authentication.AuthenticationMocker._
+import nl.knaw.dans.easy.deposit.authentication.AuthenticationProvider
+import nl.knaw.dans.easy.deposit.docs.DepositInfo
+import nl.knaw.dans.easy.deposit.{ EasyDepositApiApp, TestSupportFixture }
+import nl.knaw.dans.lib.error._
+import org.eclipse.jetty.http.HttpStatus._
+import org.scalamock.scalatest.MockFactory
+import org.scalatra.test.scalatest.ScalatraSuite
+
+class NotFoundSpec extends TestSupportFixture with ServletFixture with ScalatraSuite with MockFactory {
```
rvanheest: Rather than collecting these tests here, I would rather put them with the other tests for the specific servlet. Is there a special reason for putting them in a separate file?

jo-pol: Most servlet test are about one servlet, so they are grouped by themes.

src/test/resources/manual-test/datasetmetadata-from-ui-all.json
```diff
@@ -2,7 +2,7 @@
   "identifiers": [
     {
       "scheme": "id-type:DOI",
-      "value": "doi:10.17632/DANS.6wg5xccnjd.1"
+      "value": "10.17632/DANS.6wg5xccnjd.1"
```
jo-pol: @rvanheest note that this is a change on test data copied from the ui module.

src/test/scala/nl.knaw.dans.easy.deposit/docs/DDMSpec.scala
```diff
@@ -447,7 +443,7 @@ class DDMSpec extends TestSupportFixture with DdmBehavior {
         <ddm:accessRights>OPEN_ACCESS</ddm:accessRights>
       </ddm:profile>
       <ddm:dcmiMetadata>
-        <dcterms:identifier xsi:type="id-type:DOI">doi:10.17632/DANS.6wg5xccnjd.1</dcterms:identifier>
+        <dcterms:identifier xsi:type="id-type:DOI">10.17632/DANS.6wg5xccnjd.1</dcterms:identifier>
```
jo-pol: @rvanheest 
I don't get the prefix from `curl -i -X POST 'http://deasy.dans.knaw.nl:20140/create?type=doi'`
and there is also no prefix when downloading an XML from (for example) https://easy.dans.knaw.nl/ui/datasets/id/easy-dataset:112326/tab/1
The api.yml is not specific on this point.

PR #84, opened 2018-11-29
added tests

stephanHogenboom:
Fixes EASY-EASY-1818

#### When applied it will
* have additional unit tests


src/main/scala/nl.knaw.dans.easy.deposit/servlets/UserServlet.scala
```diff
@@ -43,6 +43,5 @@ class UserServlet(app: EasyDepositApiApp) extends ProtectedServlet(app) {
     case e: InvalidDocumentException => badDocResponse(e)
     case _ => notExpectedExceptionResponse(t)
   }
-}
-
```
rvanheest:
```suggestion
```

PR #88, opened 2018-12-03
EASY-1815 additional directory levels after upload

jo-pol:
Fixes EASY-1815

#### When applied it will
* have unit test that don't reproduce the comment on issue 1815
* The ui should post to https://github.com/DANS-KNAW/easy-deposit-api/blob/ba07e48549881f6952b7638a01d1fdef815c125e/src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala#L215
  It adds the name of the zip to the path , that creates the additional directory level.

      [2018-12-03 12:43:55,149] INFO  POST http://localhost:20190/deposit/e2abf521-38d6-416f-95f9-d768fc000bfb/file/2018-03-19.zip remote=0:0:0:0:0:0:0:1; params=[]; headers=[X-Forwarded-Host -> [deasy.dans.knaw.nl], Referer -> [http://deasy.dans.knaw.nl/deposit/deposit-form/e2abf521-38d6-416f-95f9-d768fc000bfb], Origin

* 

### DIFF
--- a/src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala
+++ b/src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala
@@ -192,6 +192,54 @@ class UploadSpec extends DepositServletFixture {
       ).map("data/path/to/dir/" + _)
 
     }
+    // get should show uploaded files
+    get(
+      uri = s"/deposit/$uuid/file/",
+      headers = Seq(fooBarBasicAuthHeader),
+    ) {
+      status shouldBe OK_200
+      body should include ("""{"filename":"readme.md","dirpath":"path/to/dir",""")
+      body should include ("""{"filename":"._login.html","dirpath":"path/to/dir/__MACOSX",""")
+      body should include ("""{"filename":"upload.html","dirpath":"path/to/dir",""")
+      body should include ("""{"filename":"login.html","dirpath":"path/to/dir",""")
+    }
+  }
+
+  it should "extract all ZIP to root of data dir in the bag" in {
+    File("src/test/resources/manual-test/Archive.zip").copyTo(testDir / "input" / "1.zip")
+    val uuid = createDeposit
+    val bagDir = testDir / "drafts" / "foo" / uuid.toString / "bag"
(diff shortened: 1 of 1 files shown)


src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala
```diff
@@ -192,6 +192,54 @@ class UploadSpec extends DepositServletFixture {
...
 
     }
+    // get should show uploaded files
+    get(
+      uri = s"/deposit/$uuid/file/",
+      headers = Seq(fooBarBasicAuthHeader),
+    ) {
+      status shouldBe OK_200
+      body should include ("""{"filename":"readme.md","dirpath":"path/to/dir",""")
+      body should include ("""{"filename":"._login.html","dirpath":"path/to/dir/__MACOSX",""")
```
rvanheest: I would not expect the `__MACOSX` directory (and its files) to be in this listing. Instead I would expect them to not even be saved in the bag, while extracting the zip. These files are used for the encoding process only.

jo-pol: @lindareijnhoudt @janvanmansum 
Didn't we decide in the past to store store whatever is stored in a zip file?
Might depend though on whether these files are created by the OS or by the zip-process.

jo-pol: When using the command line `unzip` the `__MACOSX` directory gets extracted.

jo-pol: When zipping with the command line `zip` we don't get extras.

jo-pol: If the ui zips, it shouldn't use the mac zipper. If the depositor zipped we might not be sure it was zipped by mac or even a mix.

rvanheest: UI doesn't zip; the user does. Same for ever other client that might use this api in the future.

PR #89, opened 2018-12-03
EASY-1826 add a journal log appender

stephanHogenboom:
Fixes EASY-1826

@DANS-KNAW/easy 

#### When applied it will
* add a journal log appender
Where should the reviewer @DANS-KNAW/easy start?


rvanheest:
@DANS-KNAW/easy for review

PR #90, opened 2018-12-04
EASY-1762 new deposit property

jo-pol:
Fixes EASY-1762

#### When applied it will
* create deposits with property `identifier.dans-doi.action = create`
* 
* 

Create a deposit, examine the properties. Details:
https://github.com/DANS-KNAW/easy-deposit-api/blob/master/src/test/resources/manual-test/readme.md

https://github.com/DANS-KNAW/easy-ingest-flow/pull/87 should not be deployed before this one

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
@@ -274,6 +274,7 @@ object DepositDir {
       addProperty("curation.performed", "no")
       addProperty("bag-store.bag-id", depositInfo.id)
       addProperty("identifier.doi.registered", "no")
+      addProperty("identifier.dans-doi.action", "create")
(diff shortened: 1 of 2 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -274,6 +274,7 @@ object DepositDir {
       addProperty("curation.performed", "no")
       addProperty("bag-store.bag-id", depositInfo.id)
       addProperty("identifier.doi.registered", "no")
```
lindareijnhoudt: could you rename this property at the same time? to `identifier.dans-doi.registered`

jo-pol: Please next time use the `[±]` button for simple suggestions like these:
```suggestion
      addProperty("identifier.dans-doi.registered", "no")
```

PR #86, opened 2018-11-30
EASY-1794 POST upload should not replace existing files 

jo-pol:
Fixes EASY-1794 POST upload should not replace existing files 

#### When applied it will
* [x] ~~test concurrent uploads: POST while PUT is busy moving files
  though the specs say that the result is undefined.~~
* 

#### Where should the reviewer @DANS-KNAW/easy start?

https://github.com/DANS-KNAW/easy-deposit-api/blob/4d0dcf3fe0a0c4daefc2e2dd97c97b4a668e8a87/src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala#L217

#### How should this be manually tested?

* on deasy: create a directory in `/var/opt/dans.knaw.nl/deposit/stage-upload/`
* redeploy or restart
* the service will not start, the last log line explains:

        ERROR Pending uploads were probably interrupted. See logs lines with 'POST /deposit/{id}/file/{dir_path}'. Please remove their directories from the staging area: /var/opt/dans.knaw.nl/deposit/stage-upload

#### Related pull requests on github
Conflict with #85

PR #91, opened 2018-12-04
EASY-1829 & EASY-1794 :  explicit content type & upload response codes

jo-pol:
Fixes
* EASY-1829 - explicit content type header when a response has a body
* EASY-1794 - #86 did not include the original objective of the issue:
  When uploading/deleting files, do not return a file listing

#### When applied it will
* have changes in the swagger documentation: OK -> no-content (when updating an existing file)
* 
* 

Is the following description adequate for upload of multiple files via either a zip or multiple body parts?
https://github.com/DANS-KNAW/easy-deposit-api/blob/b962a26bdac7b26b6f1def50cff8508f5c6d54e4/docs/api/api.yml#L425-L426

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
@@ -21,7 +21,7 @@ import java.util.UUID
 
 import nl.knaw.dans.easy.deposit.docs.DDM.SchemaNotAvailableException
 import nl.knaw.dans.easy.deposit.docs.JsonUtil.{ InvalidDocumentException, toJson }
-import nl.knaw.dans.easy.deposit.docs.{ DatasetMetadata, DepositInfo, StateInfo }
+import nl.knaw.dans.easy.deposit.docs.{ DatasetMetadata, StateInfo }
 import nl.knaw.dans.easy.deposit.logging._
 import nl.knaw.dans.easy.deposit.servlets.DepositServlet.{ BadRequestException, InvalidResourceException, ZipMustBeOnlyFileException }
 import nl.knaw.dans.easy.deposit.{ EasyDepositApiApp, _ }
@@ -63,7 +63,7 @@ class DepositServlet(app: EasyDepositApiApp)
       for {
         userId <- getUserId
         deposits <- app.getDeposits(userId)
-      } yield Ok(body = toJson(deposits))
+      } yield Ok(body = toJson(deposits), headers = Map(contentTypeJson))
     }.getOrRecover(respond)
   }
   post("/") {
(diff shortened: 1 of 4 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/servlets/package.scala
```diff
@@ -37,14 +37,17 @@ package object servlets extends DebugEnhancedLogging {
   private val post = "-compress(ed)?"
   private val contentTypeZipPattern = s"application/(($pre($post)?)|(x$post))"
 
+  val contentTypeJson: (String, String) = "content-type" -> "application/json;charset=UTF-8"
```
rvanheest: although this 'hardcoded' way seems fine, I think there is a method in Scalatra that does this. Something like `contentType = "..."`. Have you tried using that one? Or did you explicitly choose for this way?

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -178,25 +179,25 @@ class DepositServlet(app: EasyDepositApiApp)
...
     case e: InvalidDocumentException => badDocResponse(e)
-    case e: ConcurrentUploadException => Conflict(e.getMessage)
-    case e: FileAlreadyExistsException => Conflict("Conflict. The following file(s) already exist on the server: "+e.getMessage)
-    case e: InvalidDoiException => BadRequest(e.getMessage)
-    case e: BadRequestException => BadRequest(e.getMessage)
-    case e: ZipMustBeOnlyFileException => BadRequest(e.getMessage)
-    case e: NotImplementedException => NotImplemented(e.getMessage)
-    case e: SchemaNotAvailableException => InternalServerError(e.getMessage)
+    case e: ConcurrentUploadException => Conflict(e.getMessage, Map(contentTypePlainText))
+    case e: FileAlreadyExistsException => Conflict("Conflict. The following file(s) already exist on the server: " + e.getMessage, Map(contentTypePlainText))
```
rvanheest: Not sure if this is useful, but you can use interpolation here.

```suggestion
    case e: FileAlreadyExistsException => Conflict(s"Conflict. The following file(s) already exist on the server: ${ e.getMessage }", Map(contentTypePlainText))
```

jo-pol: matter of taste

src/main/scala/nl.knaw.dans.easy.deposit/servlets/package.scala
jo-pol: Current usage of this hard-coded constant:
https://github.com/DANS-KNAW/easy-deposit-api/blob/ef76d3475e8bb8c1b9f46453abce6addffdf90a7/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala#L75

Scalatra's convenience field is not convenient at all in this code pattern of this specific servlet. See the drawbacks of setting it with `before` in rejected PR #92

PR #96, opened 2018-12-10
EASY-1849: Revert "EASY-1826 add a journal log appender"

stephanHogenboom:
Reverts DANS-KNAW/easy-deposit-api#89
@DANS-KNAW/easy


lindareijnhoudt:
already done in #95

PR #98, opened 2018-12-13
simplify servlet loggers a bit

rvanheest:
removed calling certain methods twice in a row

@DANS-KNAW/easy for review

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/logging/RequestLogFormatter.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/logging/RequestLogFormatter.scala
@@ -15,6 +15,7 @@
  */
 package nl.knaw.dans.easy.deposit.logging
 
+import javax.servlet.http.HttpServletRequest
 import org.scalatra.{ MultiParams, ScalatraBase }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/logging/RequestLogFormatter.scala
```diff
@@ -51,7 +52,7 @@ trait RequestLogFormatter extends CookieFormatter {
     }
   }
 
-  private def getHeaderMap: HeaderMap = {
+  private def getHeaderMap(request: HttpServletRequest): HeaderMap = {
```
jo-pol: Why not use the implicit provided by ScalatraBase?

jo-pol: I would prefer you fixed your development environment.

jo-pol: Tried to synchronize and `mvn clean install` all projects?

rvanheest:
closing since this is replaces by #100

PR #100, opened 2018-12-14
NO JIRA ISSUE: removed calling certain methods twice in a row

jo-pol:
#### When applied it will
* not call certain methods twice in a row
* 
* 

Replaces #98


rvanheest:
What's wrong with my pull request

jo-pol:
As commented there: you added explicit arguments for requests that are available implicitly/globaly.

PR #101, opened 2018-12-17
NO JIRA. Amended config docs.

janvanmansum:
NO JIRA.

- [x] Question for @jo-pol  resolved.

#### When applied it will
* Update the style of the documentation comments in the config file.

### DIFF
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -56,7 +56,7 @@ pids.generator-service=http://localhost:20140/
 auth.cookie.expiresIn=3600
 
 #
-# Algorithm and secret key to use for JSON Web Token. Algorithm can be any of the values supported by jwt-scala, see fromString in:
+# Symmetric algorithm and secret key to use for JSON Web Token. Algorithm can be any of the values implementing JwtHmacAlgorithm, see fromString in:
 #
 # https://github.com/pauldijou/jwt-scala/blob/master/core/common/src/main/scala/JwtAlgorithm.scala
 #
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -38,17 +38,27 @@ users.ldap-url=ldap://localhost
 users.ldap-parent-entry=ou=users,ou=easy,dc=dans,dc=knaw,dc=nl
 users.ldap-user-id-attr-name=uid
 
-# Settings required to fetch user data when we no longer have its password.
+#
+# Settings required to fetch user data when we no longer have the user's password.
+#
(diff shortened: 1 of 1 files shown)


src/main/assembly/dist/cfg/application.properties
```diff
@@ -38,17 +38,27 @@ users.ldap-url=ldap://localhost
...
+# (Example: 3600 ( = 60 * 60 = is one hour ).
+#
 auth.cookie.expiresIn=3600
 
-# Pick a value starting with H, see fromString in https://github.com/pauldijou/jwt-scala/blob/master/core/common/src/main/scala/JwtAlgorithm.scala
+#
+# Algorithm and secret key to use for JSON Web Token. Algorithm can be any of the values supported by jwt-scala, see fromString in:
+#
+# https://github.com/pauldijou/jwt-scala/blob/master/core/common/src/main/scala/JwtAlgorithm.scala
+#
```
janvanmansum: @jo-pol : in the previous version of this comment it said a value starting with an H must be chosen? Why are the other values not valid?

jo-pol: Only hmac is imlpemented:
https://github.com/DANS-KNAW/easy-deposit-api/blob/5340ca35037c255aa76bdc7bff89113bb86ca711/src/main/scala/nl.knaw.dans.easy.deposit/authentication/TokenSupport.scala#L33-L44

jo-pol: The others a asymmetric algorithms and require two keys to encrypt/decrypt. We need only one key as cookies are both encrypted and decrypted on the server.

src/main/assembly/dist/cfg/application.properties
```diff
@@ -38,17 +38,27 @@ users.ldap-url=ldap://localhost
...
-# In seconds, 60*60=3600 is one hour
+#
+# Lifetime for JSON Web Token cookie (named "scentry.auth.default.user") in seconds.
+# (Example: 3600 ( = 60 * 60 = is one hour ).
+#
 auth.cookie.expiresIn=3600
 
-# Pick a value starting with H, see fromString in https://github.com/pauldijou/jwt-scala/blob/master/core/common/src/main/scala/JwtAlgorithm.scala
+#
+# Algorithm and secret key to use for JSON Web Token. Algorithm can be any of the values supported by jwt-scala, see fromString in:
```
jo-pol:
```suggestion
# Symmetric algorithm and secret key to use for JSON Web Token. Algorithm can be any of the values implementing JwtHmacAlgorithm, see fromString in:
```

src/main/assembly/dist/cfg/application.properties
janvanmansum: OK, thanks. I'll update the documentation comment. This way the future configurator (possibly you ;) will have a bit more understanding of the why of things.

PR #106, opened 2019-01-07
EASY-1878 Upgrade to jetty9

jo-pol:
Fixes EASY-1878 Upgrade to Jetty 9

#### When applied it will
* Upgrade the Jetty engine to v9
* not have two set-cookie headers on logout (v8 apparently kept just the last one)
* ~~no longer support basic authentication on AuthServlet~~
* no longer have test servlets that should be kept in sync with production code
* reduced the scope for the `mockedAuthenticationProvider` to prevent broken tests because another test _class_ had broken test
* [ ] Set parent to v3.0.0 when fix has been tested and new parent pom released.

login / out 
* [x] with the UI on deasy 
* [x] ~~with `src/test/resources/manual-test/login.html`~~

tried the following steps
* deasy-2018-12-12.box with https://github.com/DANS-KNAW/easy-dtap/commit/4af61038 (same date on the blessed head branch)
* vagrant destroy
* vagrant up --no-provision
* at this point wrong credentials on http://deasy.dans.knaw.nl/deposit/login did disable the login button but no browser-login form appeared
* ./deploy-role.sh easy-deposit-api
* compared the snapshot timestamp reported by maven with `yum list|grep deposit` on deasy
* checked `service easy-deposit-api status`
* checked with the browser tool that no cookies were stored for deasy
* invalid credentials as above
* valid credentials created a cookie and logout destroyed it

 
replaces #105


src/main/scala/nl.knaw.dans.easy.deposit/servlets/AuthServlet.scala
```diff
@@ -16,11 +16,20 @@
...
 
 import nl.knaw.dans.easy.deposit.EasyDepositApiApp
+import nl.knaw.dans.easy.deposit.authentication.AuthUser
 import nl.knaw.dans.easy.deposit.logging._
+import nl.knaw.dans.lib.error._
 import org.scalatra.NoContent
 
 class AuthServlet(app: EasyDepositApiApp) extends AbstractAuthServlet(app) {
 
+  override protected def fromSession: PartialFunction[String, AuthUser] = {
```
jo-pol: Would like to keep this method somehow in the authentication package, like registerAuthStrategies in ProtectedServlet below.

PR #109, opened 2019-01-22
EASY-1854: rename unauthenticated challenge for BasicAuth

rvanheest:
Fixes EASY-1854

#### When applied it will
* rename the challenge for unauthenticated response in BasicAuth to make sure the default browser login popup screen doesn't appear

@DANS-KNAW/easy for review

#### Steps to reproduce the original situation
* start a `deasy` machine with `easy-deposit-api` deployed (a version from before this PR)
* go to http://deasy.dans.knaw.nl/deposit/login in your favorite browser
* type an incorrect username/password combination
* click login
* **Now you should see the browser's default login popup.**

kapture 2019-01-24 at 9 16 29

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/authentication/EasyBasicAuthStrategy.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/authentication/EasyBasicAuthStrategy.scala
@@ -17,8 +17,8 @@ package nl.knaw.dans.easy.deposit.authentication
 
 import javax.servlet.http.{ HttpServletRequest, HttpServletResponse }
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
-import org.scalatra.ScalatraBase
 import org.scalatra.auth.strategy.BasicAuthStrategy
+import org.scalatra.{ ScalatraBase, Unauthorized }
 
 object EasyBasicAuthStrategy {}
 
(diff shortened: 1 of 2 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/authentication/EasyBasicAuthStrategy.scala
```diff
@@ -45,4 +45,16 @@ class EasyBasicAuthStrategy(protected override val app: ScalatraBase,
                          (implicit request: HttpServletRequest, response: HttpServletResponse): String = {
     user.id
   }
+
+  override def unauthenticated()(implicit request: HttpServletRequest, response: HttpServletResponse): Unit = {
+    val response = Unauthorized(headers = Map("WWW-Authenticate" -> challenge))
```
jo-pol: Somehow I can't reproduce the browser login window. 

Could you try with a blocked user or a user that hasn't yet its email confirmed? In those cases you might get no headers at all. If so, it might be enough to drop the headers here too.

https://github.com/DANS-KNAW/easy-deposit-api/blob/b46762e157834d4704098cd3f19f2ac2bcc5935c/src/main/scala/nl.knaw.dans.easy.deposit/authentication/AuthenticationSupport.scala#L43-L44

rvanheest: > Somehow I can't reproduce the browser login window.

I added some steps to reproduce, as well as a screencapture for you to the main description of this PR.

> Could you try with a blocked user or a user that hasn't yet its email confirmed? In those cases you might get no headers at all.

I've tried to login with a blocked user. However, it doesn't reach the lines you mentioned above. Instead, it already breaks down at the following line:

https://github.com/DANS-KNAW/easy-deposit-api/blob/93ef243f2acbc1e41048fee5670f1c0f0d5e81a3/src/main/scala/nl.knaw.dans.easy.deposit/authentication/LdapAuthentication.scala#L58

It filters on a user with active state, otherwise a `Success(None)` is returned by this function. Hence, `validate` below returns a `None`.

https://github.com/DANS-KNAW/easy-deposit-api/blob/93ef243f2acbc1e41048fee5670f1c0f0d5e81a3/src/main/scala/nl.knaw.dans.easy.deposit/authentication/EasyBasicAuthStrategy.scala#L33-L42

This results in a call to `unauthenticated()` which was adjusted in this PR.

However, if I remove the `.find(_.state == UserState.active)` mentioned above, it will move past the `validate` and reach the `toSession`, where it will `halt` with the proper `Unauthorized` call.

https://github.com/DANS-KNAW/easy-deposit-api/blob/93ef243f2acbc1e41048fee5670f1c0f0d5e81a3/src/main/scala/nl.knaw.dans.easy.deposit/authentication/AuthenticationSupport.scala#L39-L46

I'm not sure though if this is what is intended...
Further help would be greatly appreciated! This authentication package is quite difficult to grasp.

rvanheest:
@jo-pol is going to use another approach in EASY-1899. Closing this one.

PR #112, opened 2019-02-01
EASY-1747 postpone validation to submit (dates)

jo-pol:
Fixes EASY-1747 postpone validation to submit (authors only)

#### When applied it will
* no longer cause validation errors on dates when saving metadata
* friendly error messages are kept out of scope, see [EASY-1905](https://drivenbydata.atlassian.net/browse/EASY-1905) therefore submission relies on schema validation for now unless something would not be covered.
* [x] fix parseIntoValidForSubmit

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
@@ -161,17 +161,19 @@ object DDM extends SchemedXml with DebugEnhancedLogging {
   private implicit class RichElem(val elem: Elem) extends AnyVal {
     /** @param str the desired label, optionally with name space prefix */
     @throws[InvalidDocumentException]("when str is not a valid XML label (has more than one ':')")
-    def withLabel(str: String): Elem = withLabel(Some(str))
+    def withLabel(str: String): Elem = {
+      str.split(":") match {
+        case Array(label) if label.nonEmpty => elem.copy(label = label)
+        case Array(prefix, label) => elem.copy(prefix = prefix, label = label)
+        case a => throw invalidDatasetMetadataException(new IllegalArgumentException(
+          s"expecting (label) or (prefix:label); got [${ a.mkString(":") }] to adjust the <${ elem.label }> of ${ trim(elem) }"
+        ))
+      }
+    }
 
     @throws[InvalidDocumentException]("when maybeVal does not contain a valid XML label (its .toString has more than one ':')")
     def withLabel[T](maybeVal: Option[T]): Elem = {
-      maybeVal.map(_.toString.split(":")) match {
-        case Some(Array(label)) => elem.copy(label = label)
-        case Some(Array(prefix, label)) => elem.copy(prefix = prefix, label = label)
-        case a => throw invalidDatasetMetadataException(new IllegalArgumentException(
-          s"expecting (label) or (prefix:label); got [${ a.map(_.mkString(":")) }] to adjust the <${ elem.label }> of ${ trim(elem) }"
-        ))
-      }
+      withLabel(maybeVal.map(_.toString).getOrElse(""))
     }
   }
 
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
@@ -20,6 +20,7 @@ import java.net.UnknownHostException
 import nl.knaw.dans.easy.deposit.docs.DatasetMetadata._
 import nl.knaw.dans.easy.deposit.docs.JsonUtil.InvalidDocumentException
 import nl.knaw.dans.easy.deposit.docs.StringUtils._
+import nl.knaw.dans.easy.deposit.docs.dm.DateQualifier.DateQualifier
 import nl.knaw.dans.easy.deposit.docs.dm._
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import nl.knaw.dans.lib.string._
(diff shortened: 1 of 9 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
```diff
@@ -67,11 +67,11 @@ case class DatasetMetadata(private val identifiers: Option[Seq[SchemedValue]] =
                                                  else Failure(missingValue("AcceptDepositAgreement"))
 
   //// dates
-  private val specialDateQualifiers = Seq(DateQualifier.created, DateQualifier.available)
+  private val specialDateQualifiers = Seq(DateQualifier.created, DateQualifier.available).map(Some(_))
```
rvanheest: If you replace this whole section (https://github.com/DANS-KNAW/easy-deposit-api/blob/6646c78a249c10f5a37f14657c076c2d10d03bf6/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala#L70-L79) with the following, I think you're better off in the end. Now, you have to have `datesCreated` and `datesAvailable` and later check that it's at most one; even though your type still allows for multiple of these...

```scala
private val results = dates.getOrElse(Seq.empty)
  .foldLeft((Option.empty[Date], Option.empty[Date], Seq.empty[Date])) {
    // @formatter:off
    case ((_,           _,             _     ),      Date(_, _, Some(q@DateQualifier.dateSubmitted))) => throw new IllegalArgumentException(s"requirement failed: No $q allowed; got ${ toJson(dates) }")
    case ((None,        dateAvailable, others), date@Date(_, _, Some(  DateQualifier.created))      ) => (Some(date),  dateAvailable, others)
    case ((Some(_),     _,             _     ),      Date(_, _, Some(  DateQualifier.created))      ) => throw new IllegalArgumentException(s"requirement failed: At most one allowed; got ${ toJson(dates) }")
    case ((dateCreated, None,          others), date@Date(_, _, Some(  DateQualifier.available))    ) => (dateCreated, Some(date),    others)
    case ((_,           Some(_),       _     ),      Date(_, _, Some(  DateQualifier.available))    ) => throw new IllegalArgumentException(s"requirement failed: At most one allowed; got ${ toJson(dates) }")
    case ((dateCreated, dateAvailable, others), date@Date(_, _, _)                                  ) => (dateCreated, dateAvailable, others :+ date)
    // @formatter:on
  }
private val (_, _, plainDates) = results
val (datesCreated, datesAvailable, _) = results
val otherDates = plainDates :+ dateSubmitted()
```

rvanheest: Please note that with replacing this code, you can also get rid of `Date.atMostOne` and `Date.notAllowed`

jo-pol: Had to change the lines above into something like 
```
{ dm.datesAvailable.map(src => <ddm:available>{ src.value.getOrElse("") }</ddm:available>).getOrElse("") }
```
Then it still it broke a unit test as `At most one allowed` used to report just the duplicates, not all the dates. An attempt to build the string with `Seq(date,datesAvailable)` caused compilation problems.

src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
```diff
@@ -51,8 +51,8 @@ object DDM extends SchemedXml with DebugEnhancedLogging {
         { dm.titles.getNonEmpty.map(src => <dc:title xml:lang={ lang }>{ src }</dc:title>) }
         { dm.descriptions.getNonEmpty.map(src => <dcterms:description xml:lang={ lang }>{ src }</dcterms:description>) }
         { dm.creators.getNonEmpty.map(author => <dcx-dai:creatorDetails>{ details(author, lang) }</dcx-dai:creatorDetails>) }
-        { dm.datesCreated.map(src => <ddm:created>{ src.value }</ddm:created>) }
-        { dm.datesAvailable.map(src => <ddm:available>{ src.value }</ddm:available>) }
+        { dm.datesCreated.map(src => <ddm:created>{ src.value.getOrElse("") }</ddm:created>) }
+        { dm.datesAvailable.map(src => <ddm:available>{ src.value.getOrElse("") }</ddm:available>) }
```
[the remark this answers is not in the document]
jo-pol: 861ecb76 doesn't break test localy

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
rvanheest: perhaps you're right: if you use this `DatasetMetadata` object 'on save', I guess you have to accept anything, right? Validating on multiple `dateCreated` etc. only happens 'on submit'...

src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
```diff
@@ -161,22 +162,15 @@ object DDM extends SchemedXml with DebugEnhancedLogging {
...
-          s"expecting (label) or (prefix:label); got [${ a.mkString(":") }] to adjust the <${ elem.label }> of ${ trim(elem) }"
-        ))
-      }
-    }
+    def withLabel(str: String): Elem = withLabel(Some(str))
 
-    def withLabel[T](someEnum: Option[T]): Elem = {
-      someEnum.map(_.toString.split(":")) match {
+    @throws[InvalidDocumentException]("when maybeVal does not contain a valid XML label (its .toString has more than one ':')")
+    def withLabel[T](maybeVal: Option[T]): Elem = {
```
rvanheest: it would be more logical if you'd implement `def withLabel[T](maybeVal: Option[T])` in terms of `def withLabel(str: String)` rather than the other way around. Now you're packing and unpacking too many times.

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
jo-pol: I intended to validate AtMostOnce and NotAtAll at any time, but had less verbose error messages only reporting the problematic dates.

src/test/resources/logback.xml
```diff
@@ -12,7 +12,7 @@
     </root>
 
     <!-- Set log level during test with system property, e.g., mvn test -DLOG_LEVEL=debug -->
-    <logger name="nl.knaw.dans.easy" level="${LOG_LEVEL:-off}" />
-    <logger name="org.scalatest" level="info" />
+    <logger name="nl.knaw.dans.easy" level="trace"/>
```
rvanheest: no!

PR #113, opened 2019-02-05
EASY-1747 postpone validation to submit (spatial points/boxes)

jo-pol:
Fixes EASY-1747 postpone validation to submit (spatial points/boxes)

#### When applied it will
* no longer cause validation errors on spatial points/boxes when saving metadata
* [x] treat fields with just white space as None
* [x] ~~wait for the discussion on deposit-ui~~
  * conversion is beyond the scope of this issue, which is:
     _accept partial data on save, but not on submit_
  * api should apply same conversion of x/y to lat/long as bulk import, the source of this code

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
@@ -96,7 +96,7 @@ case class DatasetMetadata(private val identifiers: Option[Seq[SchemedValue]] = 
 
     for { // TODO collect errors
       _ <- Author.validate(authors)
-      _ <- Spatial.validate((spatialPoints ++ spatialBoxes).toSeq.flatten)
+      _ <- Spatial.hasMandatory((spatialPoints++ spatialBoxes).toSeq.flatten)
     } yield ()
   }
 }
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
@@ -93,7 +93,11 @@ case class DatasetMetadata(private val identifiers: Option[Seq[SchemedValue]] = 
 
   /** Validations as far as not covered by DDM schema validation. */
   private[docs] def validate(): Try[Unit] = {
-    Author.validate(authors)
+
+    for { // TODO collect errors
+      _ <- Author.validate(authors)
+      _ <- Spatial.validate((spatialPoints ++ spatialBoxes).toSeq.flatten)
+    } yield ()
   }
 }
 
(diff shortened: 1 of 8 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
```diff
@@ -93,7 +93,26 @@ case class DatasetMetadata(private val identifiers: Option[Seq[SchemedValue]] =
...
+      else {
+        Failure(new IllegalArgumentException(
+          s"Missing values for $msg: $invalid"
+        ))
+      }
+    }
+
+    for { // TODO collect errors over multiple fields, not just within the list of a field
+      _ <- Author.validate(authors)
+      // filterNot: stick to schema validation for now, error message of missingMandatory not clear
```
jo-pol: Perhaps just ignore Contributors/Creators and other fields without any value. Opinions?

For that matter perhaps refactor use of `getNonEmpty`-s in `DDM.toXml`

src/main/scala/nl.knaw.dans.easy.deposit/docs/DatasetMetadata.scala
```diff
@@ -93,7 +93,21 @@ case class DatasetMetadata(private val identifiers: Option[Seq[SchemedValue]] =
...
+        .withFilter(!_.hasMandatory)
+        .map(x => x.getClass.getSimpleName + JsonUtil.toJson(x))
+        .mkString(", ")
+      if (invalid.isEmpty) Success(())
+      else {
+        Failure(new IllegalArgumentException(s"Missing mandatory values in: $invalid"))
+      }
+    }
+
+    for { // TODO collect errors over multiple types of validation errors
```
rvanheest: you are looking for https://typelevel.org/cats/datatypes/validated.html 🤔 
we'll get to that later 😜 
keep the TODO for now!

src/main/scala/nl.knaw.dans.easy.deposit/docs/dm/Spatial.scala
```diff
@@ -23,45 +25,48 @@ object Spatial {
...
+                        y: Option[String],
+                       ) extends SchemedSpatial {
+  private val sx: String = x.getOrElse("")
+  private val sy: String = y.getOrElse("")
   lazy val pos: String = srsName match {
-    case Spatial.RD_SRS_NAME => s"$x $y"
-    case Spatial.DEGREES_SRS_NAME => s"$y $x"
-    case _ => s"$y $x"
+    case Spatial.RD_SRS_NAME => s"$sx $sy"
+    case Spatial.DEGREES_SRS_NAME => s"$sy $sx"
```
rvanheest: There is a chance this might change. I'm doubting whether I should give you the lat/lon in the correct form from the UI, rather than you doing the conversion here...
I'll discuss this with @lindareijnhoudt tomorrow

PR #115, opened 2019-02-21
EASY-1962 fix bag-info.txt

jo-pol:
Fixes EASY-1962 fix bag-info.txt

#### When applied it will
* dropped user-account
* added created
* 

#### How should this be manually tested?

* build and deploy on deasy
* Submit a deposit with the ui
* Work around https://drivenbydata.atlassian.net/browse/EASY-1932
 `sudo chgrp -R deposits /var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox/` 
 `sudo chmod -R 755 /var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox/` 
 `sudo service easy-ingest-flow restart`

ingest-flow now only complains about `metadata/message-from-depositor.txt` but that is to be fixed with https://drivenbydata.atlassian.net/browse/EASY-1961

#### Related pull requests on github
repo                       | PR                | note
-------------------------- | ----------------- | ----
easy-                      | [PR#](PRlink)     | ..

PR #116, opened 2019-02-21
EASY-1932 file rights before moving to ingest inbox

jo-pol:
Fixes EASY-1932 file rights before moving to ingest inbox

#### When applied it will
* add group "deposits" to files
* give write access to group
* as a consequence of the above bullets, a deposit with the ui (alias PUT /deposit/{id}/state SUBMITTED) will no longer cause a warning about not writeable files
* [x] decide on configuration parameter when updating dtap 
* [x] test that replaces the dropped one with a non existing group
* [x] rename actual directories to mimic the renamed configuration variables

workaround in #115 would no longer be needed 
will still cause validation errors unless the other known issues are fixed

* https://github.com/DANS-KNAW/easy-dtap/pull/301 
* the draft #119 has already merge conflicts with this pull request

### DIFF
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -21,6 +21,7 @@ deposits.drafts=/var/opt/dans.knaw.nl/tmp/easy-deposit/drafts
 deposits.stage=/var/opt/dans.knaw.nl/tmp/easy-deposit/stage
 deposits.submit-to=/var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox
 
+# file rights that allow easy-ingest-flow to process deposits from its inbox
 deposit.permissions.access=rwxrwx---
 deposit.permissions.group=deposits
 
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -22,7 +22,6 @@ deposits.stage=/var/opt/dans.knaw.nl/tmp/easy-deposit/stage
 deposits.submit-to=/var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox
 
 # file rights that allow easy-ingest-flow to process deposits from its inbox
-deposit.permissions.access=rwxrwx---
 deposit.permissions.group=deposits
 
 #
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -21,7 +21,7 @@ deposits.drafts=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/drafts
 deposits.stage=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage
 deposits.submit-to=/var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox
 
-# file rights that allow easy-ingest-flow to process deposits from its inbox
+# The group to assign to the files and directories making up the deposit, before moving it to submit-to.
 deposit.permissions.group=deposits
 
 #
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -6,19 +6,22 @@
 daemon.http.port=20190
 
 #
-# Directories for managing the deposits. The lifecycle of a deposit is:
+# Directories for managing the deposits.
 #
-# 1.   Created in the drafts directory.
-# 2.   On submit: copied to stage and then that copy is moved to the submit-to directory.
-# 3a)  If archived successfully the bag in the stage is moved to submit-to
-# 3b)  If the deposit is rejected, the depositor can reopen the draft, modify it and resubmit it.
+# - stage-zips:       is used for uploads of zipped directories or complete deposits.
+# - drafts:           contains the draft deposits; single files are uploaded here directly,
+#                     unzipped directories and deposits are moved here from stage-zips.
+# - stage-for-submit: the completed draft is copied here when the user submits it.
+# - submit-to:        the target directory to which the copy in stage-for-submit is moved for further processing.
 #
-# stage and submit-to must be on a single mount to allow moving from the first to the latter
-# stage.upload and draft must be on a single mount to allow moving zip-content into the draft bag
+# N.B. stage-zips and drafts must be located on the same partition in order to support atomic moving.
+# The same holds for stage-for-submit and submit-to.
 #
-deposits.stage.upload=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage-upload
+# Note that the stage-* directories only have temporary content, so the service should clear them automatically.
+#
+deposits.stage-zips=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage-upload
 deposits.drafts=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/drafts
-deposits.stage=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage
+deposits.stage-for-submit=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage
 deposits.submit-to=/var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox
 
 # The group to assign to the files and directories making up the deposit, before moving it to submit-to.
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -12,12 +12,15 @@ daemon.http.port=20190
 # - drafts:           contains the draft deposits; single files are uploaded here directly,
 #                     unzipped directories and deposits are moved here from stage-zips.
 # - stage-for-submit: the completed draft is copied here when the user submits it.
-# - submit-to:        the target directory to which the copy in stage-for-submit is moved for further processing.
+# - submit-to:        the deposit copy in stage-for-submit is moved here on successful completion
+#                     of a submit request.  As moving is an atomic action this prevents further
+#                     processing of incomplete deposits.
 #
 # N.B. stage-zips and drafts must be located on the same partition in order to support atomic moving.
 # The same holds for stage-for-submit and submit-to.
 #
-# Note that the stage-* directories only have temporary content, so the service should clear them automatically.
+# Note that the stage-* directories only have temporary content, so the service should clear them
+# automatically on completion of a request.
 #
 deposits.stage-zips=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage-zips
 deposits.drafts=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/drafts
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -19,9 +19,9 @@ daemon.http.port=20190
 #
 # Note that the stage-* directories only have temporary content, so the service should clear them automatically.
 #
-deposits.stage-zips=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage-upload
+deposits.stage-zips=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage-zips
 deposits.drafts=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/drafts
-deposits.stage-for-submit=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage
+deposits.stage-for-submit=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage-for-submit
 deposits.submit-to=/var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox
 
 # The group to assign to the files and directories making up the deposit, before moving it to submit-to.
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -10,11 +10,11 @@ daemon.http.port=20190
 #
 # 1.   Created in the drafts directory.
 # 2.   On submit: copied to stage and then that copy is moved to the submit-to directory.
-# 3a)  If archived successfully the bag in the draft-copy is deleted
+# 3a)  If archived successfully the bag in the stage is moved to submit-to
 # 3b)  If the deposit is rejected, the depositor can reopen the draft, modify it and resubmit it.
 #
-# stage and submit-to must be on a single mount to allow moving
-# stage.upload and draft must be on a single mount to allow moving
+# stage and submit-to must be on a single mount to allow moving from the first to the latter
+# stage.upload and draft must be on a single mount to allow moving zip-content into the draft bag
 #
 deposits.stage.upload=/var/opt/dans.knaw.nl/tmp/easy-deposit/stage-upload
 deposits.drafts=/var/opt/dans.knaw.nl/tmp/easy-deposit/drafts
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -21,6 +21,9 @@ deposits.drafts=/var/opt/dans.knaw.nl/tmp/easy-deposit/drafts
 deposits.stage=/var/opt/dans.knaw.nl/tmp/easy-deposit/stage
 deposits.submit-to=/var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox
 
+deposit.permissions.access=rwxrwx---
+deposit.permissions.group=deposits
+
 #
 # Settings for user authentication. The LDAP directory at ldap-url is used to check the credentials
 # of the user by trying to access the entry:
--- a/src/main/assembly/dist/cfg/application.properties
+++ b/src/main/assembly/dist/cfg/application.properties
@@ -1,4 +1,8 @@
-### General settings ###
+#
+# EASY deposit-api configuration file
+#
+# Please, modify the configuration below to match your environment. Also, delete and/or modify
+# this description.
 
 #
 # Port that the HTTP-service listens on.
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -15,9 +15,8 @@
  */
 package nl.knaw.dans.easy.deposit
 
-import java.io.{ IOException, InputStream }
+import java.io.InputStream
 import java.net.URI
-import java.nio.file.attribute.{ GroupPrincipal, UserPrincipalNotFoundException }
 import java.nio.file.{ FileAlreadyExistsException, Path }
 import java.util.UUID
 
@@ -31,7 +30,6 @@ import nl.knaw.dans.lib.error._
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.configuration.PropertiesConfiguration
 
-import scala.util.control.NonFatal
 import scala.util.{ Failure, Success, Try }
 
 class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLogging
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -15,8 +15,9 @@
  */
 package nl.knaw.dans.easy.deposit
 
-import java.io.InputStream
+import java.io.{ IOException, InputStream }
 import java.net.URI
+import java.nio.file.attribute.UserPrincipalNotFoundException
 import java.nio.file.{ FileAlreadyExistsException, Path }
 import java.util.UUID
 
@@ -30,6 +31,7 @@ import nl.knaw.dans.lib.error._
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.configuration.PropertiesConfiguration
 
+import scala.util.control.NonFatal
 import scala.util.{ Failure, Success, Try }
 
 class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLogging
@@ -70,11 +72,19 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
     dir
   }
   private val draftsDir = getConfiguredDirectory("deposits.drafts")
-  private lazy val submitter = new Submitter(
-    stagingBaseDir = getConfiguredDirectory("deposits.stage"),
-    submitToBaseDir = getConfiguredDirectory("deposits.submit-to"),
-    group = configuration.properties.getString("deposit.permissions.group"),
-  )
+  private lazy val submitter = {
+    val stagingDir = getConfiguredDirectory("deposits.stage")
+    val group = configuration.properties.getString("deposit.permissions.group")
+    val submitToDir = getConfiguredDirectory("deposits.submit-to")
+    val principal = Try {
+      stagingDir.path.getFileSystem.getUserPrincipalLookupService.lookupPrincipalByGroupName(group)
+    }.getOrRecover {
+      case e: UserPrincipalNotFoundException => throw new IOException(s"Group $group could not be found", e)
+      case e: UnsupportedOperationException => throw new IOException("Not on a POSIX supported file system", e)
+      case NonFatal(e) => throw new IOException(s"unexpected error occured on $stagingDir", e)
+    }
+    new Submitter(stagingDir, submitToDir, principal)
+  }
 
   private def getConfiguredDirectory(key: String): File = {
     val dir = File(configuration.properties.getString(key))
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -60,7 +60,7 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
   }
 
   private val uploadStagingDir = {
-    val dir = getConfiguredDirectory("deposits.stage.upload")
+    val dir = getConfiguredDirectory("deposits.stage-zips")
     logger.info(s"Uploads are staged in $dir")
     if (dir.nonEmpty) {
       val msg = s"Pending uploads were probably interrupted. See logs lines with 'POST /deposit/{id}/file/{dir_path}'. Please remove their directories from the staging area: $dir"
@@ -72,7 +72,7 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
   private val draftsDir = getConfiguredDirectory("deposits.drafts")
 
   private val submitter = new Submitter(
-    getConfiguredDirectory("deposits.stage"),
+    getConfiguredDirectory("deposits.stage-for-submit"),
     getConfiguredDirectory("deposits.submit-to"),
     configuration.properties.getString("deposit.permissions.group"),
   )
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -17,7 +17,7 @@ package nl.knaw.dans.easy.deposit
 
 import java.io.{ IOException, InputStream }
 import java.net.URI
-import java.nio.file.attribute.UserPrincipalNotFoundException
+import java.nio.file.attribute.{ GroupPrincipal, UserPrincipalNotFoundException }
 import java.nio.file.{ FileAlreadyExistsException, Path }
 import java.util.UUID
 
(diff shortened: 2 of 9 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -85,9 +86,15 @@ class Submitter(stagingBaseDir: File,
     // EASY-1464 3.3.7 checksums
     _ <- samePayloadManifestEntries(stageBag, draftBag)
     // EASY-1464 step 3.3.9 Move copy to submit-to area
+    _ = stageBag.baseDir.parent.list.foreach(setRights(_))
     _ = stageBag.baseDir.parent.moveTo(submitDir)
   } yield ()
 
+  private def setRights(file: File) = {
+    file.addPermission(PosixFilePermission.GROUP_WRITE)
+    file.setGroup("deposits")
```
jo-pol: Causes `java.nio.file.attribute.UserPrincipalNotFoundException` in unit tests.

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -85,9 +86,15 @@ class Submitter(stagingBaseDir: File,
     // EASY-1464 3.3.7 checksums
     _ <- samePayloadManifestEntries(stageBag, draftBag)
     // EASY-1464 step 3.3.9 Move copy to submit-to area
+    _ = stageBag.baseDir.parent.list.foreach(setRights(_))
     _ = stageBag.baseDir.parent.moveTo(submitDir)
   } yield ()
 
+  private def setRights(file: File) = {
+    file.addPermission(PosixFilePermission.GROUP_WRITE)
```
jo-pol: When deployed this causes: `java.io.IOException: 'owner' parameter can't be a group`

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -85,9 +86,15 @@ class Submitter(stagingBaseDir: File,
     // EASY-1464 3.3.7 checksums
     _ <- samePayloadManifestEntries(stageBag, draftBag)
     // EASY-1464 step 3.3.9 Move copy to submit-to area
+    _ = stageBag.baseDir.parent.list.foreach(setRights(_))
```
rvanheest: shouldn't there be a comment referencing EASY-1932? or not?

rvanheest:
```suggestion
    _ = stageBag.baseDir.parent.list.foreach(setRights)
```

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -85,9 +86,15 @@ class Submitter(stagingBaseDir: File,
...
     _ = stageBag.baseDir.parent.moveTo(submitDir)
   } yield ()
 
+  private def setRights(file: File) = {
```
rvanheest:
```suggestion
  private def setRights(file: File): Unit = {
```

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
rvanheest:
```suggestion
      .setGroup("deposits")
```

rvanheest: is it OK to hard-code this group?

rvanheest: I vaguely remember having seen similar things in `easy-split-multi-deposit`. Maybe you could take a look at: https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/73189001217c2bf31b487eb8356f76ea4e9ffc31/src/main/scala/nl.knaw.dans.easy.multideposit/actions/SetDepositPermissions.scala#L72-L90

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -84,10 +93,34 @@ class Submitter(stagingBaseDir: File,
     _ <- isValid(stageBag)
     // EASY-1464 3.3.7 checksums
     _ <- samePayloadManifestEntries(stageBag, draftBag)
+    _ <- stageBag.baseDir.parent.walk().toStream.map(setRights).find(_.isFailure).getOrElse(Success(()))
```
rvanheest: Are you 100% sure that **in all circumstances** the `Iterator` produced by `.walk()` is closed? I currently don't have the mental capacity to proof this to myself 😞 
Concretely: if this fails (you chose 'fail-fast'), is the underlying connection to the file system closed?

rvanheest: _Thought from later in this review:_

Perhaps I'm a bit conservative here, but could you take another look at [the code in `easy-split-multi-deposit`](https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/73189001217c2bf31b487eb8356f76ea4e9ffc31/src/main/scala/nl.knaw.dans.easy.multideposit/actions/SetDepositPermissions.scala#L72-L90) and the context in which this is used. As you see we use a `SimpleFileVisitor` from `java.nio.file` here in combination with `Files.walkFileTree`. I'm wondering if this wouldn't be safer to use instead, especially in light of the resource closing mentioned in my previous comment.

src/test/scala/nl.knaw.dans.easy.deposit/SubmitterSpec.scala
```diff
@@ -97,41 +128,41 @@ class SubmitterSpec extends TestSupportFixture with MockFactory {
...
     assume(DDM.triedSchema.isAvailable)
-    new Submitter(testDir / "staged", testDir / "submitted").submit(depositDir) should matchPattern {
+    new Submitter(testDir / "staged", testDir / "submitted", "rwxrwx---", userGroup).submit(depositDir) should matchPattern {
       case Failure(e)
         if e.getMessage == s"staged and draft bag [${ bag.baseDir.parent }] have different payload manifest elements: (Set((data/file.txt,$checksum)),Set((data/file.txt,${ checksum }xxx)))" =>
     }
   }
 
-  private def getBagDir(depositDir: DepositDir) = getBag(depositDir).baseDir
+  private def addDoiToDepositProperties(bag: DansBag) = {
```
rvanheest:
```suggestion
  private def addDoiToDepositProperties(bag: DansBag): Unit = {
```

src/test/scala/nl.knaw.dans.easy.deposit/TestSupportFixture.scala
```diff
@@ -30,12 +30,29 @@ import org.scalamock.scalatest.MockFactory
...
 
+import scala.util.Properties
+
 trait TestSupportFixture extends FlatSpec with Matchers with Inside with BeforeAndAfterEach with MockFactory {
   implicit def existenceOfFile[FILE <: better.files.File]: Existence[FILE] = _.exists
 
   lazy val testDir: File = currentWorkingDirectory / "target" / "test" / getClass.getSimpleName
   lazy val uuid: UUID = UUID.randomUUID()
 
+  lazy val (user, userGroup, unrelatedGroup) = {
```
rvanheest: looks like you copied this from [`easy-split-multi-deposit`](https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/ea7c2dc3d6a78cacd779d2efcb2d799e399e22cf/src/test/scala/nl.knaw.dans.easy.multideposit/actions/SetDepositPermissionsSpec.scala#L28-L41)... Please state so in a comment for future reference!

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -71,7 +71,9 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
   private val draftsDir = getConfiguredDirectory("deposits.drafts")
   private lazy val submitter = new Submitter(
     stagingBaseDir = getConfiguredDirectory("deposits.stage"),
-    submitToBaseDir = getConfiguredDirectory("deposits.submit-to")
+    submitToBaseDir = getConfiguredDirectory("deposits.submit-to"),
+    permissions = configuration.properties.getString("deposit.permissions.access"),
```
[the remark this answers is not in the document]
jo-pol: for dtap it might be clearer to have the same names in deposit-api as in split-multi-deposit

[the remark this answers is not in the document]
rvanheest: I guess in your `easy-dtap` implementation you could have the same property used in both the configurations for split-multi-deposit and deposit-api, indeed. Good point!

However, my point is with the naming of these properties here in deposit-api. They should be clear as to what kind of permissions are set.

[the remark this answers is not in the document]
jo-pol: Perhaps we need renames for
https://github.com/DANS-KNAW/easy-dtap/blob/733923e168f192c301700a097178eb55f1e5e995/provisioning/group_vars/easy_servers.yml#L344-L345
```
easy_split_multi_deposit_output_permissions: "rwxrwx---"
easy_split_multi_deposit_output_group: "deposits"
```
These values are not used but hard coded for `easy-split-muliti-deposit` and `easy-creation-work`.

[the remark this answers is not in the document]
rvanheest: well, not rename those, but put the values in `shared_xxx`, just as we do for other shared values.

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
jo-pol: Under the hood `walk` uses a java stream. From [stack-overflow](https://stackoverflow.com/questions/38339440/how-to-convert-a-java-stream-to-a-scala-stream): "both ways will fully consume the Java Stream".

Not yet sure this applies also to my `tostream`, which by the way is not required to compile. If with toStream the underlying java stream is also completely consumed, the fail-fast approach only prevents further calls to `setRights` after a failure.

jo-pol: ps: what's under the hood
https://github.com/pathikrit/better-files/blob/4f2d9b5adf12753af6b1487bfed87e8be1d30073/core/src/main/scala/better/files/File.scala#L783-L784
The Iterator is actually a java.util.stream.Stream[Path]

rvanheest: almost... 😃 what you don't see on GitHub is that there is an implicit conversion going on here. Open the same file in IntelliJ and turn on the implicit conversions (https://www.jetbrains.com/help/idea/edit-scala-code.html#scala_hints)

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -33,7 +39,10 @@ import scala.util.{ Failure, Success, Try }
  * @param submitToBaseDir the directory to which the staged copy must be moved.
  */
 class Submitter(stagingBaseDir: File,
-                submitToBaseDir: File) {
+                submitToBaseDir: File,
+                group: String,
+               ) {
+  private val groupPrincipal = stagingBaseDir.path.getFileSystem.getUserPrincipalLookupService.lookupPrincipalByGroupName(group)
```
jo-pol: Making this variable lazy would postpone the error from service start to the first request, but is that desired behaviour?

rvanheest: I think we should fail with this ASAP, preferably on startup! However, isn't this already the case with the `lazy val` for the `Submitter itself?

Perhaps you should do something similar to `getConfiguredDirectory` with this `groupPrinciple` in `EasyDepositApiApp` and give the `GroupPrinciple` as an argument to `Submitter`.

src/main/assembly/dist/cfg/application.properties
```diff
@@ -10,17 +10,20 @@ daemon.http.port=20190
 #
 # 1.   Created in the drafts directory.
 # 2.   On submit: copied to stage and then that copy is moved to the submit-to directory.
-# 3a)  If archived successfully the bag in the draft-copy is deleted
+# 3a)  If archived successfully the bag in the stage is moved to submit-to
```
janvanmansum: @jo-pol: Why? Doesn't make sense to me.

src/main/assembly/dist/cfg/application.properties
```diff
@@ -10,17 +10,20 @@ daemon.http.port=20190
...
 # 3b)  If the deposit is rejected, the depositor can reopen the draft, modify it and resubmit it.
 #
-# stage and submit-to must be on a single mount to allow moving
-# stage.upload and draft must be on a single mount to allow moving
+# stage and submit-to must be on a single mount to allow moving from the first to the latter
+# stage.upload and draft must be on a single mount to allow moving zip-content into the draft bag
```
janvanmansum: What is `stage.upload`?

src/main/assembly/dist/cfg/application.properties
```diff
@@ -10,17 +10,20 @@ daemon.http.port=20190
...
 #
 deposits.stage.upload=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage-upload
 deposits.drafts=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/drafts
 deposits.stage=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage
 deposits.submit-to=/var/opt/dans.knaw.nl/tmp/easy-ingest-flow-inbox
 
+# file rights that allow easy-ingest-flow to process deposits from its inbox
```
janvanmansum:
```suggestion
# The group to assign to the files and directories making up the deposit, before moving it to submit-to.
```

Let's keep this self-contained and not bring in easy-ingest-flow in the description.

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -84,10 +94,38 @@ class Submitter(stagingBaseDir: File,
...
+    trace(path)
+    // EASY-1932, variant of https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/73189001217c2bf31b487eb8356f76ea4e9ffc31/src/main/scala/nl.knaw.dans.easy.multideposit/actions/SetDepositPermissions.scala#L72-L90
+    File(path).addPermission(GROUP_WRITE)
+    // tried File(path).setGroup(groupPrincipal) but it causes "java.io.IOException: 'owner' parameter can't be a group"
+    Files.getFileAttributeView(
+      path,
+      classOf[PosixFileAttributeView],
+      LinkOption.NOFOLLOW_LINKS,
+    ).setGroup(groupPrincipal)
+  }.recoverWith {
```
janvanmansum: Could all these exceptions just be called `e` or will that result in a naming conflict? I agree that a long name is not necessary, but names like `upnf`, though fun to pronounce, don't add any information either.

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -70,11 +72,19 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
-  private lazy val submitter = new Submitter(
-    stagingBaseDir = getConfiguredDirectory("deposits.stage"),
-    submitToBaseDir = getConfiguredDirectory("deposits.submit-to"),
-    group = configuration.properties.getString("deposit.permissions.group"),
-  )
+  private lazy val submitter = {
+    val stagingDir = getConfiguredDirectory("deposits.stage")
+    val group = configuration.properties.getString("deposit.permissions.group")
+    val submitToDir = getConfiguredDirectory("deposits.submit-to")
+    val principal = Try {
```
rvanheest: extract method?

src/main/assembly/dist/cfg/application.properties
jo-pol: Perhaps replace the life cycle approach with something like:

```
# drafts
#   - each user/uuid directory contains a bag + deposit.properties
# stage.upload
#   - contains temporary directories with zip content during processing of upload requests
#   - the temporary directories are prefixed with a user-id and uuid
#   - on failure the content is deleted
#   - on success the content is added as payload to a draft bag
# stage
#   - each uuid directory holds copies of draft deposits during processing a submit request
#   - on failure the copy is deleted
#   - on success the copy is moved to submit-to
# submitted
#   - each uuid directory contains a copy of a bag + deposit.properties that is ready for curation/ingest
```

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -70,10 +72,18 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
   }
   private val draftsDir = getConfiguredDirectory("deposits.drafts")
-  private lazy val submitter = new Submitter(
-    stagingBaseDir = getConfiguredDirectory("deposits.stage"),
-    submitToBaseDir = getConfiguredDirectory("deposits.submit-to")
-  )
+  private val stagingDir = getConfiguredDirectory("deposits.stage")
+  private val submitToDir = getConfiguredDirectory("deposits.submit-to")
+
+  private val group = configuration.properties.getString("deposit.permissions.group")
```
rvanheest: could you do an extract method of these two variables into something like `private val principal = getPrinciple("deposit.permissions.group")` to have it more look like the `getConfiguredDirectory` calls above?

src/main/assembly/dist/cfg/application.properties
janvanmansum: @jo-pol as discussed off-line we'll go for the following. Note that some keys need to be renamed:
```
#
# Directories for managing the deposits.
#
# - stage-zips: is used for uploads of zipped directories or complete deposits.
# - drafts: contains the draft deposits; single files are uploaded here directly, unzipped directories and deposits
#     are moved here from stage-zips.
# - stage-for-submit: the completed draft is copied here when the user submits it.
# - submit-to: the target directory to which the copy in stage-for-submit is moved for further processing.
# 
# N.B. stage-zips and drafts must be located on the same partition in order to support atomic moving. The same 
# holds for stage-for-submit and submit-to.
#
# Note that the stage-* directories only have temporary content, so the service should clear them automatically.
#
deposits.stage-zips=
deposits.drafts
deposits.stage-for-submit
deposits.submit-to


```

src/main/assembly/dist/cfg/application.properties
```diff
@@ -6,21 +6,27 @@
...
+# Directories for managing the deposits.
 #
-# 1.   Created in the drafts directory.
-# 2.   On submit: copied to stage and then that copy is moved to the submit-to directory.
-# 3a)  If archived successfully the bag in the draft-copy is deleted
-# 3b)  If the deposit is rejected, the depositor can reopen the draft, modify it and resubmit it.
+# - stage-zips:       is used for uploads of zipped directories or complete deposits.
+# - drafts:           contains the draft deposits; single files are uploaded here directly,
+#                     unzipped directories and deposits are moved here from stage-zips.
+# - stage-for-submit: the completed draft is copied here when the user submits it.
```
rvanheest: perhaps add a couple of words on why a '_completed draft_' must first be moved to `stage-for-submit`, before it is copied to the `submit-to` directory. That wasn't clear to me on first reading.

src/main/assembly/dist/cfg/application.properties
```diff
@@ -6,21 +6,27 @@
...
+# - submit-to:        the target directory to which the copy in stage-for-submit is moved for further processing.
```
rvanheest:
```suggestion
# - submit-to:        the target directory to which the copy in stage-for-submit is atomically moved for further processing.
```

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -70,9 +70,11 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
   }
   private val draftsDir = getConfiguredDirectory("deposits.drafts")
-  private lazy val submitter = new Submitter(
-    stagingBaseDir = getConfiguredDirectory("deposits.stage"),
-    submitToBaseDir = getConfiguredDirectory("deposits.submit-to")
+
+  private val submitter = new Submitter(
+    getConfiguredDirectory("deposits.stage-for-submit"),
+    getConfiguredDirectory("deposits.submit-to"),
+    configuration.properties.getString("deposit.permissions.group"),
```
rvanheest: I see you moved '_finding the `GroupPrinciple`_' back to `Submitter`. Why? Shouldn't an incorrect group result in a `ConfigurationException`, just as with `getConfiguredDirectory`?

Also, wouldn't moving '_finding the `GroupPrinciple`_' to here and passing a `GroupPrinciple` to `Submitter` be more beneficial for testing/mocking purposes?

src/main/assembly/dist/cfg/application.properties
rvanheest: side note, I see this documentation is duplicated in `src/test/resources`. Please keep them in sync!
If necessary, add a test for that (although I can imagine you've already done so 😉)

src/test/resources/debug-config/application.properties
```diff
@@ -6,21 +6,27 @@
...
-deposits.stage.upload=data/stage-upload
+# Note that the stage-* directories only have temporary content, so the service should clear them automatically.
+#
+deposits.stage-zips=data/stage-zips
 deposits.drafts=data/drafts
-deposits.stage=data/stage
+deposits.stage-for-submit=data/stage-for-submit
 deposits.submit-to=data/easy-ingest-flow-inbox
 
+# file rights that allow easy-ingest-flow to process deposits from its inbox
```
rvanheest: documentation not in sync with the `src/main/assembly` variant...

src/test/scala/nl.knaw.dans.easy.deposit/SubmitterSpec.scala
```diff
@@ -37,17 +40,37 @@ class SubmitterSpec extends TestSupportFixture with MockFactory {
   private val doi = datasetMetadata.doi
     .getOrElse(fail("could not get DOI from test input"))
 
-  "submit" should "write all files" in {
+  "lazy constructor" should "not be called by EasyDepositApiApp if the configured group does not exist" in {
+    val props = minimalAppConfig.properties
+    props.setProperty("deposit.permissions.group", "not-existing-group")
+
+    Try(new EasyDepositApiApp(new Configuration("", props))) should matchPattern {
```
rvanheest: I think you can do something like:

```scala
the [IOException] thrownBy(new EasyDepositApiApp(new Configuration("", props))) should have message "Group not-existing-group could not be found"
```

might be more descriptive of what you're actually testing here

src/test/scala/nl.knaw.dans.easy.deposit/SubmitterSpec.scala
```diff
@@ -97,41 +118,41 @@ class SubmitterSpec extends TestSupportFixture with MockFactory {
...
     assume(DDM.triedSchema.isAvailable)
-    new Submitter(testDir / "staged", testDir / "submitted").submit(depositDir) should matchPattern {
+    new Submitter(testDir / "stage-for-submit", testDir / "submitted", userGroup).submit(depositDir) should matchPattern {
       case Failure(e)
         if e.getMessage == s"staged and draft bag [${ bag.baseDir.parent }] have different payload manifest elements: (Set((data/file.txt,$checksum)),Set((data/file.txt,${ checksum }xxx)))" =>
     }
   }
 
-  private def getBagDir(depositDir: DepositDir) = getBag(depositDir).baseDir
+  private def addDoiToDepositProperties(bag: DansBag): File = {
```
rvanheest: Yes, according to `better-files` it returns `File`, but since this method _appends_ to a file, you should return `Unit`.

```suggestion
  private def addDoiToDepositProperties(bag: DansBag): Unit = {
```

src/main/assembly/dist/cfg/application.properties
jo-pol: DebugConfigSpec checks for the same files in both config directories and for the same keys in both application.properties files. The values are different. Don't know how to check for comments.

Perhaps reduce comment in debug-config to a reference to `dist/config`?

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
jo-pol: The submitter used to be lazily created by the `App` class. I didn't see the reason why and dropped the lazyness. It prevented values in `App` that were otherwise not used. Actually testing/mocking became easier.

src/main/assembly/dist/cfg/application.properties
rvanheest: How about reading the file for the 'comment test' as lines, filtering the ones that start with `#` and comparing the resulting output.

```scala
val srcFile = srcConfig.lines.filter(_ startsWith "# ").toList
val testFile = testConfig.lines.filter(_ startsWith "# ").toList

srcFile should contain theSameElementsInOrderAs testFile
```

jo-pol: If you really want that comment check: create an issue for all 16 modules that have a `DebugConfigSpec`. You may come up with more checks.

src/test/scala/nl.knaw.dans.easy.deposit/SubmitterSpec.scala
```diff
@@ -37,17 +40,36 @@ class SubmitterSpec extends TestSupportFixture with MockFactory {
   private val doi = datasetMetadata.doi
     .getOrElse(fail("could not get DOI from test input"))
 
-  "submit" should "write all files" in {
+  "lazy constructor" should "not be called by EasyDepositApiApp if the configured group does not exist" in {
```
rvanheest: description seems incorrect???

src/test/scala/nl.knaw.dans.easy.deposit/SubmitterSpec.scala
```diff
@@ -15,20 +15,23 @@
...
 
 import better.files.StringOps
+import nl.knaw.dans.bag.DansBag
 import nl.knaw.dans.easy.deposit.docs.StateInfo.State
 import nl.knaw.dans.easy.deposit.docs._
 import nl.knaw.dans.lib.error._
 import org.scalamock.scalatest.MockFactory
 
-import scala.util.{ Failure, Success }
+import scala.util.{ Failure, Success, Try }
```
rvanheest: unused import?

src/main/assembly/dist/cfg/application.properties
janvanmansum: > side note, I see this documentation is duplicated in `src/test/resources`. Please keep them in sync!
> If necessary, add a test for that (although I can imagine you've already done so 😉)

I think there is little value in keeping the comments in `src/test/resources`. I'd say: just leave them out of those. They are just for development testing and developers should know where to find them in `src/main/assembly`.

jo-pol: I'm happy with dropping the comment duplication.

The archteype has the following header in both `dist/config` as in `debug-config`.
```
#
# EASY Module configuration file
#
# Please, modify the configuration below to match your environment. Also, delete and/or modify
# this description.
```
* Shouldn't this comment header be dropped in `debug-config` of the archetype too?
* Should the archetype replace `Module` in `dist/config` with the module name?
* Should we synchronise the chosen convention in all our scala modules?

src/main/assembly/dist/cfg/application.properties
```diff
@@ -1,26 +1,39 @@
...
+#                     processing of incomplete deposits.
 #
-# stage and submit-to must be on a single mount to allow moving
-# stage.upload and draft must be on a single mount to allow moving
+# N.B. stage-zips and drafts must be located on the same partition in order to support atomic moving.
+# The same holds for stage-for-submit and submit-to.
 #
-deposits.stage.upload=/var/opt/dans.knaw.nl/tmp/easy-deposit-api/stage-upload
+# Note that the stage-* directories only have temporary content, so the service should clear them
+# automatically on completion of a request.
```
jo-pol: @lindareijnhoudt 
Is this comment block clear to you now?

src/main/assembly/dist/cfg/application.properties
```diff
@@ -1,26 +1,39 @@
...
+# - drafts:           contains the draft deposits; single files are uploaded here directly,
+#                     unzipped directories and deposits are moved here from stage-zips.
+# - stage-for-submit: the completed draft is copied here when the user submits it.
+# - submit-to:        the deposit copy in stage-for-submit is moved here on successful completion
+#                     of a submit request.  As moving is an atomic action this prevents further
+#                     processing of incomplete deposits.
 #
-# stage and submit-to must be on a single mount to allow moving
-# stage.upload and draft must be on a single mount to allow moving
+# N.B. stage-zips and drafts must be located on the same partition in order to support atomic moving.
```
rvanheest: > `stage-zips` and `drafts` must be located on the same partition in order to support atomic moving. The same holds for `stage-for-submit` and `submit-to`.

Do you know whether it is possible to verify these requirements on application startup? Not sure about it, but it seems quite important to have that tested on startup.

jo-pol: created an issue for this:
https://drivenbydata.atlassian.net/browse/EASY-2001

PR #124, opened 2019-03-25
EASY-2013 document how lat/long is mapped to x/y

jo-pol:
Fixes EASY-2013 document how lat/long is mapped to x/y

#### When applied it will
* 
* 
* 

### DIFF
--- a/docs/api/api.yml
+++ b/docs/api/api.yml
@@ -1059,7 +1059,7 @@ components:
       properties:
         scheme:
           type: string
-          description: "Possible values: degrees, RD."
+          description: "Recommended values: http://www.opengis.net/def/crs/EPSG/0/4326 (degrees), http://www.opengis.net/def/crs/EPSG/0/28992 (RD)."
         x:
           type: string
           format: "number"
--- a/docs/api/api.yml
+++ b/docs/api/api.yml
@@ -1059,12 +1059,15 @@ components:
       properties:
         scheme:
           type: string
+          description: "Possible values: degrees, RD."
         x:
           type: string
           format: "number"
+          description: "In case of degrees: latitude, range from -90 to 90."
         y:
           type: string
           format: "number"
+          description: "In case of degrees: longitude, range from -180 to 180."
 
     # Note: it is not possible to provide examples in combination with a reference to this
     # schema. If you need examples, you'll have to spell out the complete type.
--- a/docs/api/api.yml
+++ b/docs/api/api.yml
@@ -1063,11 +1063,11 @@ components:
         x:
           type: string
           format: "number"
-          description: "In case of degrees: latitude, range from -90 to 90."
+          description: "In case of RD: range from -7000 to 300000. In case of degrees: latitude, range from -90 to 90."
         y:
           type: string
           format: "number"
-          description: "In case of degrees: longitude, range from -180 to 180."
+          description: "In case of RD: range from 289000 to 629000. In case of degrees: longitude, range from -180 to 180."
 
     # Note: it is not possible to provide examples in combination with a reference to this
     # schema. If you need examples, you'll have to spell out the complete type.
--- a/docs/api/api.yml
+++ b/docs/api/api.yml
@@ -1059,7 +1059,7 @@ components:
       properties:
         scheme:
           type: string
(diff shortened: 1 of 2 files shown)


docs/api/api.yml
```diff
@@ -1059,12 +1059,15 @@ components:
...
           type: string
+          description: "Possible values: degrees, RD."
         x:
           type: string
           format: "number"
+          description: "In case of RD: range from -7000 to 300000. In case of degrees: latitude, range from -90 to 90."
         y:
           type: string
           format: "number"
+          description: "In case of RD: range from 289000 to 629000. In case of degrees: longitude, range from -180 to 180."
```
jo-pol: When simplifying the valid area for RD to a rectangle, we have an area of 11x11 km of overlap between possible x/y values. However, the actual [valid area](https://nl.wikipedia.org/wiki/Rijksdriehoeksco%C3%B6rdinaten#Geldigheidsgebied) is a polygon.

docs/api/api.yml
```diff
@@ -1059,12 +1059,15 @@ components:
       properties:
         scheme:
           type: string
+          description: "Possible values: degrees, RD."
```
rvanheest: seems incorrect. The following are the correct values:

* `RD` -> `http://www.opengis.net/def/crs/EPSG/0/28992`
* `degrees` -> `http://www.opengis.net/def/crs/EPSG/0/4326`

@jo-pol please verify that this is reflected in the code!

jo-pol: Actually, the api and DDM.xsd allow everything. The api maps the values `RD` and `degrees` (case sensitive) to the URI's for the DDM.

https://github.com/DANS-KNAW/easy-deposit-api/blob/5b20f651586305350e754bf8fe868943798b200a/src/main/scala/nl.knaw.dans.easy.deposit/docs/dm/Spatial.scala#L29-L35
https://github.com/DANS-KNAW/easy-deposit-api/blob/5b20f651586305350e754bf8fe868943798b200a/src/test/scala/nl.knaw.dans.easy.deposit/docs/DDMSpec.scala#L310-L317

docs/api/api.yml
```diff
@@ -1059,12 +1059,15 @@ components:
       properties:
         scheme:
           type: string
+          description: "Recommended values: http://www.opengis.net/def/crs/EPSG/0/4326 (degrees), http://www.opengis.net/def/crs/EPSG/0/28992 (RD)."
```
rvanheest:
```suggestion
          description: "Recommended values: 'http://www.opengis.net/def/crs/EPSG/0/4326' (degrees), 'http://www.opengis.net/def/crs/EPSG/0/28992' (RD)."
```

docs/api/api.yml
jo-pol: Now that I looked into the code: since the api doesn't validate, should we really document it here? Did not try whether ingest flow adds validation to the `ddm.xsd`

rvanheest: @lindareijnhoudt what do you think? Should we document these values here?

Please note, as far as I can tell, the value for `scheme` is **never** validated in `easy-deposit-api`, `easy-ingest-flow` or any of the other services related to ingest. Hence, you can basically put whatever in here...

lindareijnhoudt: yes, the idea there being that you are allowed to use whatever coordinate-system you want. only if it is provided in one of these two, we can understand what is meant, and propagate accordingly.

so yes, it should be documented here, to tell that these are understood by EASY

jo-pol: @rvanheest Note that possible has been changed into recommended.

docs/api/api.yml
rvanheest: @lindareijnhoudt do we want to improve the validation of the RD coordinates, given the rules explained in the link above. If we do, we need to do that in all services (`easy-deposit-ui`, `easy-split-multi-deposit`, `easy-validate-dans-bag` (???), `easy-ddm` (???)).

rvanheest: I'll merge this PR for now, though this question has not yet been resolved. But if we want this more precise validation, we'll make a different issue for that anyway.

lindareijnhoudt: it makes sense to have it in easy-deposit-ui and in split-multi-deposit, as that are entry points. for the sword-input we make the assumption the front-office takes care of it. 
doesn't the opengis gml already validate this? (i guess not the RD)

PR #127, opened 2019-03-28
Easy-1948 check status before deposit updates

stephanHogenboom:
Fixes EASY-1948

#### When applied it will
*  validate the state of deposit before changing files of the deposit

- [x] more unit tests?
- [ ]   Test on DEASY
- [ ] verify the check is done everywhere it is needed
- [ ] update swagger

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/StateInfo.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/StateInfo.scala
@@ -23,7 +23,7 @@ import scala.collection.Seq
 import scala.util.{ Failure, Success, Try }
 
 case class StateInfo(state: State, stateDescription: String) {
-  def isDeletable: Try[Unit] = {
+  def canDelete: Try[Unit] = {
     if (StateInfo.deletableStates.contains(state)) Success(())
     else Failure(new IllegalStateException(s"Deposit has state $state, can only delete deposits with one of the states: ${ StateInfo.deletableStates.mkString(", ") }"))
   }
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/StateInfo.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/StateInfo.scala
@@ -27,10 +27,16 @@ case class StateInfo(state: State, stateDescription: String) {
     if (StateInfo.deletableStates.contains(state)) Success(())
     else Failure(new IllegalStateException(s"Deposit has state $state, can only delete deposits with one of the states: ${ StateInfo.deletableStates.mkString(", ") }"))
   }
+
+  def canUpdate: Try[Unit] = {
+    if (StateInfo.updatableStates.contains(state)) Success(())
+    else Failure(new IllegalStateException(s"Deposit has state $state, can only update deposits with one of the states: ${ StateInfo.updatableStates.mkString(", ") }"))
+  }
 }
 
 object StateInfo {
   val deletableStates: Seq[State] = Seq(State.draft, State.archived, State.rejected)
+  val updatableStates: Seq[State] = Seq(State.draft, State.rejected)
 
   object State extends Enumeration {
     type State = Value
--- a/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
@@ -105,8 +105,7 @@ class DepositServlet(app: EasyDepositApiApp)
         _ <- app.checkDoi(user.id, uuid, datasetMetadata)
         _ <- app.writeDataMetadataToDeposit(datasetMetadata, user.id, uuid)
       } yield NoContent()
-    }.doIfFailure{ case t: Throwable => println(t.getMessage)}
-      .getOrRecover(respond)
+    }.getOrRecover(respond)
       .logResponse
   }
   get("/:uuid/state") {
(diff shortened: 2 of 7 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/docs/StateInfo.scala
```diff
@@ -27,10 +27,16 @@ case class StateInfo(state: State, stateDescription: String) {
     if (StateInfo.deletableStates.contains(state)) Success(())
     else Failure(new IllegalStateException(s"Deposit has state $state, can only delete deposits with one of the states: ${ StateInfo.deletableStates.mkString(", ") }"))
   }
+
+  def canUpdate: Try[Unit] = {
```
rvanheest: Make this name equivalent to the one above.

However, I would suggest `canDelete` there, rather than changing it here. @jo-pol what do you think?

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -249,8 +254,15 @@ class DepositServlet(app: EasyDepositApiApp)
     }
   }
 
-  private def getRequestBodyAsManagedInputStream = {
-    Try { managed(request.getInputStream) }
+  private def getRequestBodyAsManagedInputStream: Try[ManagedResource[ServletInputStream]] = Try {
```
rvanheest: The `Try` seems unnecessary here?

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -249,8 +254,15 @@ class DepositServlet(app: EasyDepositApiApp)
...
+    managed(request.getInputStream)
+  }
+
+  private def canUpdate(uuid: UUID): Try[Unit] = {
+    for {userId <- getUserId
+         deposit <- app.getDepositState(userId, uuid)
```
rvanheest: This isn't the `deposit` but the `state`, right?

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
```diff
@@ -151,6 +153,7 @@ class DepositServlet(app: EasyDepositApiApp)
         uuid <- getUUID
         path <- getPath
         _ <- isMultipart
+        _ <- canUpdate(uuid)
```
jo-pol: The `canDelete` (isDeletable) is called by the `DepositDir` class. This servlet class should only unwrap the http-request and wrap up the results in a response, it should not contain execution logic. Must admit I violated this rule above by calling `app.checkDoi` which should be moved into `writeDataMetadataToDeposit`. In this particular case I still have my doubts I did get this right.

A side note probably a bit beyond the scope of this issue: I made a bit of a mess in the past from reading de deposit properties and getting the status out of it. The problem was avoiding reading again and again and not getting into trouble with changes.

src/main/scala/nl.knaw.dans.easy.deposit/servlets/DepositServlet.scala
stephanHogenboom: so true -.-

src/test/scala/nl.knaw.dans.easy.deposit/docs/StateInfoSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+    new StateInfo(StateInfo.State.inProgress, "IN_PROGRESS").isDeletable should matchPattern {
+      case Failure(ise: IllegalStateException) if ise.getMessage == s"Deposit has state IN_PROGRESS, can only delete deposits with one of the states: ${ StateInfo.deletableStates.mkString(", ") }" =>
+    }
+  }
+
+  "canUpdate" should "succeed if the state is rejected" in {
+    new StateInfo(StateInfo.State.rejected, "rejected").canUpdate shouldBe a[Success[_]]
+  }
+
+  it should "succeed if the the state if draft" in {
```
rvanheest: you use `the` twice in this description as well as below

src/test/scala/nl.knaw.dans.easy.deposit/servlets/HappyRoutesSpec.scala
```diff
@@ -174,4 +175,40 @@ class HappyRoutesSpec extends TestSupportFixture with ServletFixture with Scalat
...
+  private def expectPutMetaDataToFail(stateInfo: StateInfo): Unit = {
+    put(
+      uri = s"/deposit/$uuid/metadata",
+      headers = Seq(fooBarBasicAuthHeader)
+    ) {
+      status shouldBe FORBIDDEN_403
+      body shouldBe s"""Deposit has state ${ stateInfo.state }, can only update deposits with one of the states: DRAFT, REJECTED"""
+    }
+  }
+
```
rvanheest: newlines 🤔

src/test/scala/nl.knaw.dans.easy.deposit/servlets/HappyRoutesSpec.scala
```diff
@@ -174,4 +175,40 @@ class HappyRoutesSpec extends TestSupportFixture with ServletFixture with Scalat
...
+  it should "reject the deposit if the deposit has in progress as state" in {
+    val stateInfo = StateInfo(StateInfo.State.inProgress, "in progress")
+    authMocker.expectsUserFooBar
+    (mockedApp.getDepositState(_: String, _: UUID)) expects("foo", uuid) returning
+      Success(stateInfo)
+    expectPutMetaDataToFail(stateInfo)
+  }
+
+  private def expectPutMetaDataToFail(stateInfo: StateInfo): Unit = {
```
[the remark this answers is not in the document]
stephanHogenboom: removed test, since they are no longer part of the servlet's behaviour but the app's

PR #128, opened 2019-03-29
EASY-2001 implemented same mount check at start-up

jo-pol:
Fixes EASY-2001 implemented same mount check at start-up

#### When applied it will
* not start the service when staged directories are configured on other mounts than their partners (the directories were the temporary content should move to atomically).
* 
* 

* build and deploy
* change the location of the `stage-`* folders in `application.properties` and create these directories at another mount than their partners (were the temporary content is moved to) on deasy:
  `sudo mkdir /home/stage-for-submit /home/stage-zips`
  `sudo chown easy-deposit-api /home/stage-*`
  `sudo service easy-deposit-api restart`
* the service won't restart and logs an error in `easy-deposit-api.err`

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -63,9 +63,9 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
     val dir = getConfiguredDirectory("deposits.stage-zips")
     logger.info(s"Uploads are staged in $dir")
     if (dir.nonEmpty) {
-      val msg = s"Possibly pending or interrupted uploads in the staging area: $dir. See log lines with 'POST /deposit/{id}/file/{dir_path}'."
-      logger.error(msg)
-      throw new FileAlreadyExistsException(msg)
+      throw new FileAlreadyExistsException(
+        s"Possibly pending or interrupted uploads in the staging area: $dir. See log lines with 'POST /deposit/{id}/file/{dir_path}'."
+      )
     }
     dir
   }
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -15,8 +15,9 @@
  */
 package nl.knaw.dans.easy.deposit
 
-import java.io.InputStream
+import java.io.{ IOException, InputStream }
 import java.net.URI
+import java.nio.file.spi.FileSystemProvider
 import java.nio.file.{ FileAlreadyExistsException, Path }
 import java.util.UUID
 
@@ -63,14 +64,16 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
     val dir = getConfiguredDirectory("deposits.stage-zips")
     logger.info(s"Uploads are staged in $dir")
     if (dir.nonEmpty) {
+      // TODO move to Validation and/or new Uploader class for DepositServlet post("/:uuid/file/*")
       throw new FileAlreadyExistsException(
-        s"Possibly pending or interrupted uploads in the staging area: $dir. See log lines with 'POST /deposit/{id}/file/{dir_path}'."
+        s"Upload staging area [$dir] should be empty unless force shutdown during an upload request. Check logging related to 'POST /deposit/{id}/file/{dir_path}'."
       )
     }
     dir
   }
   private val draftsDir: File = getConfiguredDirectory("deposits.drafts")
-  sameMounts(uploadStagingDir.fileSystem.provider(), uploadStagingDir, draftsDir)
+  private val provider: FileSystemProvider = uploadStagingDir.fileSystem.provider()
+  Validation.sameMounts(provider, uploadStagingDir, draftsDir)
 
   private val submitter = new Submitter(
     getConfiguredDirectory("deposits.stage-for-submit"),
@@ -78,8 +81,13 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
     configuration.properties.getString("deposit.permissions.group"),
   )
 
-  private def getConfiguredDirectory(key: String): File = {
-    val dir = File(configuration.properties.getString(key))
+  @throws[ConfigurationException]("when no existing readable directory is configured")
+  def getConfiguredDirectory(key: String): File = {
+    // TODO move to Validation?
+    val str = Try{configuration.properties.getString(key)}.recoverWith { case e =>
+      throw ConfigurationException(s"$e")
+    }.unsafeGetOrThrow
+    val dir = File(str)
 
     if (!dir.exists) throw ConfigurationException(s"Configured directory '$key' does not exist: $dir")
     if (!dir.isDirectory) throw ConfigurationException(s"Configured directory '$key' is a regular file: $dir")
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -84,9 +84,8 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
   @throws[ConfigurationException]("when no existing readable directory is configured")
   private def getConfiguredDirectory(key: String): File = {
     // TODO move to Validation?
-    val str = Try { configuration.properties.getString(key) }.recoverWith { case e =>
-      throw ConfigurationException(s"$e")
-    }.unsafeGetOrThrow
+    val str = Option(configuration.properties.getString(key))
+      .getOrElse(throw ConfigurationException(s"No configuration value for $key"))
     val dir = File(str)
 
     if (!dir.exists) throw ConfigurationException(s"Configured directory '$key' does not exist: $dir")
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -63,13 +63,14 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
     val dir = getConfiguredDirectory("deposits.stage-zips")
     logger.info(s"Uploads are staged in $dir")
     if (dir.nonEmpty) {
-      val msg = s"Pending uploads were probably interrupted. See logs lines with 'POST /deposit/{id}/file/{dir_path}'. Please remove their directories from the staging area: $dir"
+      val msg = s"Possibly pending or interrupted uploads the in staging area: $dir. See log lines with 'POST /deposit/{id}/file/{dir_path}'."
       logger.error(msg)
       throw new FileAlreadyExistsException(msg)
     }
     dir
   }
-  private val draftsDir = getConfiguredDirectory("deposits.drafts")
+  private val draftsDir: File = getConfiguredDirectory("deposits.drafts")
+  sameMounts(uploadStagingDir.fileSystem.provider(), uploadStagingDir, draftsDir)
 
   private val submitter = new Submitter(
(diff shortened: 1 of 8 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -63,13 +63,14 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
     val dir = getConfiguredDirectory("deposits.stage-zips")
     logger.info(s"Uploads are staged in $dir")
     if (dir.nonEmpty) {
-      val msg = s"Pending uploads were probably interrupted. See logs lines with 'POST /deposit/{id}/file/{dir_path}'. Please remove their directories from the staging area: $dir"
+      val msg = s"Possibly pending or interrupted uploads the in staging area: $dir. See log lines with 'POST /deposit/{id}/file/{dir_path}'."
```
rvanheest:
```suggestion
      val msg = s"Possibly pending or interrupted uploads in the staging area: $dir. See log lines with 'POST /deposit/{id}/file/{dir_path}'."
```

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -63,13 +63,14 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
       logger.error(msg)
```
rvanheest: Should you still do `logger.error` here? Given that you throw an exception here as well, doesn't this crash the application and write that to the `*.err` log file already? No this error log probably ends up in two places.

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -103,10 +106,11 @@ class Submitter(stagingBaseDir: File,
...
     // EASY-1464 3.3.7 checksums
     _ <- samePayloadManifestEntries(stageBag, draftBag)
-    _ <- setRightsRecursively(stageBag.baseDir.parent)
+    draftDepositDir = stageBag.baseDir.parent
+    _ <- setRightsRecursively(draftDepositDir)
     // EASY-1464 step 3.3.9 Move copy to submit-to area
-    _ = logger.info(s"moving ${stageBag.baseDir.parent} to $submitDir")
-    _ = stageBag.baseDir.parent.moveTo(submitDir)
+    _ = logger.info(s"moving $draftDepositDir to $submitDir")
+    _ = srcProvider.move(draftDepositDir.path, submitDir.path, StandardCopyOption.ATOMIC_MOVE)
```
rvanheest: good to see the extraction of `stageBag.baseDir.parent`.
I'm wondering though whether you need to change the `moveTo` to a `srcProvider.move` here. Since atomic moves are possible (which you checked on startup), do you still need to be explicit about it here? Can't you just simply use `draftDepositDir.moveTo(submitDir, overwrite = false)`?

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -53,6 +54,8 @@ class Submitter(stagingBaseDir: File,
       case NonFatal(e) => throw new IOException(s"unexpected error occured on $stagingBaseDir", e)
     }
   }
+  val srcProvider: FileSystemProvider = stagingBaseDir.fileSystem.provider()
```
rvanheest: how is this `stagingBaseDir.fileSystem.provider()` different from the `stagingBaseDir.path.getFileSystem` above?

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
jo-pol: better safe than sorry with a double check than rely on configuration checks.

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
jo-pol: overwrite false is a good idea, otherwise better safe than sorry with a double check than rely on configuration checks.

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -63,13 +63,14 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
     val dir = getConfiguredDirectory("deposits.stage-zips")
     logger.info(s"Uploads are staged in $dir")
     if (dir.nonEmpty) {
-      val msg = s"Pending uploads were probably interrupted. See logs lines with 'POST /deposit/{id}/file/{dir_path}'. Please remove their directories from the staging area: $dir"
-      logger.error(msg)
-      throw new FileAlreadyExistsException(msg)
+      throw new FileAlreadyExistsException(
+        s"Possibly pending or interrupted uploads in the staging area: $dir. See log lines with 'POST /deposit/{id}/file/{dir_path}'."
```
rvanheest: The sentence starting with `See log lines` doesn't make any sense to me, yet.

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
rvanheest: **Reminder to future me:**

Look at why we were able to pull this off in [`easy-split-multi-deposit`](https://github.com/DANS-KNAW/easy-split-multi-deposit/blob/628bfe965c1e69658b94dd314de318eb3310ca1a/src/main/scala/nl.knaw.dans.easy.multideposit/actions/MoveDepositToOutputDir.scala#L35), but why we can't seem to do the same here.

src/main/scala/nl.knaw.dans.easy.deposit/package.scala
```diff
@@ -71,4 +74,19 @@ package object deposit {
...
+  def sameMounts(srcProvider: FileSystemProvider, src: File, target: File): Unit = {
+    val fileName = "same-mount-check"
+    val tempFile = src / fileName
+    val movedTempFile = target / fileName
+    Try {
+      if (tempFile.notExists) tempFile.createFile()
+      srcProvider.move(tempFile.path, movedTempFile.path, StandardCopyOption.ATOMIC_MOVE)
+      movedTempFile.delete()
+    }.recoverWith { case e =>
+      tempFile.delete()
```
rvanheest: `tempFile.delete()` is just a side-effect. Use `doIfFailure` instead.

src/main/scala/nl.knaw.dans.easy.deposit/package.scala
```diff
@@ -71,4 +74,19 @@ package object deposit {
         .getOrElse(onSuccess)
     }
   }
+
+  @throws[IOException]("when files can not be moved atomically from src to target")
+  def sameMounts(srcProvider: FileSystemProvider, src: File, target: File): Unit = {
```
rvanheest: I'm not yet a big fan of putting this validation in the package object. Could you think of another (dedicated) place for these kinds of startup validations?

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
jo-pol: The directory should be empty after completing request. The most important reason for left overs could be force shutdown during upload. Logged request without response can conform that situation. Otherwise we might have other problems.

src/main/scala/nl.knaw.dans.easy.deposit/package.scala
jo-pol: It is called from constructors so I need to throw. `doIfFailure` returns a try.

src/main/scala/nl.knaw.dans.easy.deposit/Validation.scala
```diff
@@ -0,0 +1,26 @@
...
+import java.io.IOException
+import java.nio.file.StandardCopyOption
+import java.nio.file.spi.FileSystemProvider
+
+import better.files.File
+import nl.knaw.dans.lib.error._
+
+import scala.util.Try
+
+object Validation {
```
rvanheest: `StartupValidation`?

src/test/scala/nl.knaw.dans.easy.deposit/servlets/HappyRoutesSpec.scala
```diff
@@ -15,9 +15,11 @@
  */
 package nl.knaw.dans.easy.deposit.servlets
 
+import java.nio.file.spi.FileSystemProvider
```
rvanheest: unused imports?

src/test/scala/nl.knaw.dans.easy.deposit/SubmitterSpec.scala
```diff
@@ -105,8 +103,7 @@ class SubmitterSpec extends TestSupportFixture with MockFactory {
     val depositDir = createDeposit(datasetMetadata.copy(messageForDataManager = None))
     addDoiToDepositProperties(getBag(depositDir))
 
-    new Submitter(testDir / "stage-for-submit", testDir / "submitted", userGroup)
-      .submit(depositDir) should matchPattern { case Success(()) => }
+    createSubmitter(userGroup).submit(depositDir) should matchPattern { case Success(()) => }
```
rvanheest:
```suggestion
    createSubmitter(userGroup).submit(depositDir) shouldBe a[Success[_]]
```

src/test/scala/nl.knaw.dans.easy.deposit/SubmitterSpec.scala
```diff
@@ -78,8 +77,7 @@ class SubmitterSpec extends TestSupportFixture with MockFactory {
     }
 
     // the test
-    new Submitter(testDir / "stage-for-submit", testDir / "submitted", userGroup)
-      .submit(depositDir) should matchPattern { case Success(()) => }
+    createSubmitter(userGroup).submit(depositDir) should matchPattern { case Success(()) => }
```
rvanheest:
```suggestion
    createSubmitter(userGroup).submit(depositDir) shouldBe a[Success[_]]
```

src/main/scala/nl.knaw.dans.easy.deposit/package.scala
```diff
@@ -15,17 +15,20 @@
...
+import better.files.{ File, StringOps }
 import nl.knaw.dans.bag.DansBag
 import nl.knaw.dans.easy.deposit.docs.StateInfo.State.State
+import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 
 import scala.util.{ Failure, Try }
 import scala.xml._
 
-package object deposit {
+package object deposit extends DebugEnhancedLogging {
```
rvanheest: no debug logging necessary anymore, as well as unused imports

```suggestion
package object deposit {
```

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -63,22 +64,30 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
     getConfiguredDirectory("deposits.submit-to"),
     configuration.properties.getString("deposit.permissions.group"),
   )
 
-  private def getConfiguredDirectory(key: String): File = {
-    val dir = File(configuration.properties.getString(key))
+  @throws[ConfigurationException]("when no existing readable directory is configured")
+  def getConfiguredDirectory(key: String): File = {
+    // TODO move to Validation?
+    val str = Try{configuration.properties.getString(key)}.recoverWith { case e =>
```
rvanheest: `getString` returns `null` of the property is not present. `Try` doesn't help you here...
maybe use `Option(...).getOrElse { throw ... }`?

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
rvanheest: I've been looking at this tonight, but I can't figure out why I don't specify an **atomic move** in `easy-split-multi-deposit`, but do seem to get the correct behavior. I'll try to take another (final) look at it tomorrow with @janvanmansum.

PR #129, opened 2019-03-29
Easy-2033 check status before deposit updates

stephanHogenboom:
Fixes EASY-2033

- [x] test on deasy
- [x] update swagger

#### Related pull requests on github
repo                       | PR                | note
-------------------------- | ----------------- | ----
easy-                      | [PR#](PRlink)     | ..

PR #131, opened 2019-04-01
EASY-1949 + EASY-1950 checks on PUT file (no zip, no file overwriting directory)

jo-pol:
Fixes EASY-1949 + EASY-1950 checks on PUT file (no zip, no file overwriting directory)

#### When applied it will
* Fix EASY-1949 reject zip files for PUT file requests
* Fix EASY-1950 reject a file that would overwrite a directory
* 

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -262,8 +262,8 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
    * @param id          the deposit ID
    * @param path        the path of the file to (over)write, relative to the content base directory
    * @param is          the input stream to write from
-   * @param contentType the contentType of the stream, it originates from an HTTP request header
-   *                    which is optional by nature but mandatory in this context
+   * @param contentType the contentType of the stream, mandatory in this context
+   *                    but optional by the nature of an HTTP request header
    * @return `true` if a new file was created, `false` otherwise
    */
   def writeDepositFile(is: => InputStream, user: String, id: UUID, path: Path, contentType: Option[String]): Try[Boolean] = {
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -26,6 +26,8 @@ import nl.knaw.dans.easy.deposit.PidRequesterComponent.PidRequester
 import nl.knaw.dans.easy.deposit.authentication.LdapAuthentication
 import nl.knaw.dans.easy.deposit.docs.StateInfo.State
 import nl.knaw.dans.easy.deposit.docs.{ DatasetMetadata, DepositInfo, StateInfo }
+import nl.knaw.dans.easy.deposit.servlets.DepositServlet.{ BadRequestException, ConflictException }
+import nl.knaw.dans.easy.deposit.servlets.contentTypeZipPattern
 import nl.knaw.dans.lib.error._
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.configuration.PropertiesConfiguration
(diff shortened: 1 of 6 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -262,9 +264,13 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
    * @param is   the input stream to write from
    * @return `true` if a new file was created, `false` otherwise
    */
-  def writeDepositFile(is: => InputStream, user: String, id: UUID, path: Path): Try[Boolean] = {
+  def writeDepositFile(is: => InputStream, user: String, id: UUID, path: Path, contentType: String): Try[Boolean] = {
     for {
+      _ <- if (Option(contentType).toSeq.exists(!_.trim.matches(contentTypeZipPattern))) Success(())
+           else Failure(BadRequestException("Content-Type must not be application/zip."))
       dataFiles <- getDataFiles(user, id)
+      _ <- if (!(dataFiles.bag / "data" / path.toString).isDirectory) Success(())
```
jo-pol: Don't loose this or #129 (canUpdate) when fixing merge conflicts

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -262,9 +264,13 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
    * @param is   the input stream to write from
```
rvanheest: don't forget to expand the docs with this new parameter

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -262,9 +264,13 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
    * @return `true` if a new file was created, `false` otherwise
    */
-  def writeDepositFile(is: => InputStream, user: String, id: UUID, path: Path): Try[Boolean] = {
+  def writeDepositFile(is: => InputStream, user: String, id: UUID, path: Path, contentType: String): Try[Boolean] = {
```
rvanheest: If you make `contentType` of type `Option[String]` rather than allowing it to be `null`, you make this code much easier to read!

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -262,9 +264,13 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
     for {
+      _ <- if (Option(contentType).toSeq.exists(!_.trim.matches(contentTypeZipPattern))) Success(())
```
rvanheest: I don't think `.toSeq` is necessary here. `exists` is also defined on `Option`

```suggestion
      _ <- if (Option(contentType).exists(!_.trim.matches(contentTypeZipPattern))) Success(())
```

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
rvanheest: if you swap the if-else branches, your code becomes a little more easy to read.

Or perhaps even a better idea: extract method for both these if-else constructs that you added?

src/main/scala/nl.knaw.dans.easy.deposit/servlets/package.scala
```diff
@@ -42,10 +42,10 @@ import scala.util.{ Failure, Success, Try }
  */
 package object servlets extends DebugEnhancedLogging {
 
-  private val extensionZipPattern = ".*.g?z(ip)?"
+  val extensionZipPattern = ".*.g?z(ip)?"
```
rvanheest: this is of course a regex. Let me see if I understand this one correctly. The following should all match **the latter part** of this pattern:

* ".gzip"
* ".gz"
* ".zip"
* ".z"

especially the last one is one that I'm doubting this pattern should match. Am I overlooking something?

jo-pol: It might have been unintentionally but now that you make me look again... We recently discussed `tar` but there is also `rar` and and a lot more exotic variations.

It is more in the scope of [EASY-2028](https://drivenbydata.atlassian.net/browse/EASY-2028) to check which types we support.

https://www.file-extensions.org/z-file-extension
https://www.file-extensions.org/filetype/extension/name/archive-and-compressed-files

jo-pol: The documentation only talks about the content-type. Should we look at the extension anyway? Again more in the scope of EASY-2028.

https://dans-knaw.github.io/easy-deposit-api/api.html#/directory/post_deposit__id__file__dir_path_

jo-pol: A unit test shows the .z was intentional
https://github.com/DANS-KNAW/easy-deposit-api/blob/7f03c5d174a6970b089db7f90b8efb18c90908a3/src/test/scala/nl.knaw.dans.easy.deposit/servlets/RichFileItemsSpec.scala#L52-L69

PR #132, opened 2019-04-03
Easy 2028  allow nested zips

stephanHogenboom:
Fixes Easy-2028  allow nested zips

#### When applied it will
* posting nested zips as datasets 

- [x]  test on deasy
- [x] update specs?

#### Related pull requests on github
repo                       | PR                | note
-------------------------- | ----------------- | ----
easy-deposit-api                      | [PR#129](https://github.com/DANS-KNAW/easy-deposit-api/pull/129)     | ..

PR #136, opened 2019-04-09
EASY-1947 multipart not allowed for PUT file requests

jo-pol:
Fixes EASY-1947 PUT file with Content-Type: multipart/form-data results in corrupt upload

#### When applied it will
* fix an oversight in documentation of nested zips in #132
* document/check no multipart is PUT file request

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -277,7 +277,7 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
    */
   def writeDepositFile(is: => InputStream, user: String, id: UUID, path: Path, contentType: Option[String]): Try[Boolean] = {
     for {
-      _ <- contentTypeAnythingButZip(contentType)
+      _ <- contentTypeAnythingBut(contentType)
       dataFiles <- getDataFiles(user, id)
       _ <- canUpdate(user, id)
       _ <- pathNotADirectory(path, dataFiles)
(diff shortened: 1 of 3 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -293,10 +293,13 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
   }
 
-  private def contentTypeAnythingButZip(contentType: Option[String]): Try[Unit] = {
-    if (contentType.exists(str => str.trim.nonEmpty && !str.trim.matches(contentTypeZipPattern)))
-      Success(())
-    else Failure(InvalidContentTypeException(contentType, "must not be a zip."))
+  private def contentTypeAnythingBut(contentType: Option[String]): Try[Unit] = {
+    if (contentType.map(_.trim.toLowerCase).exists(str =>
+      str.nonEmpty &&
+      !str.matches(contentTypeZipPattern) &&
```
rvanheest: 1. would it be wise to swap the `matches` and `startsWith` tests: the latter is faster and can more easily fail
2. formatting please 😉

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -293,10 +293,13 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
     else Success(())
   }
 
-  private def contentTypeAnythingButZip(contentType: Option[String]): Try[Unit] = {
-    if (contentType.exists(str => str.trim.nonEmpty && !str.trim.matches(contentTypeZipPattern)))
-      Success(())
-    else Failure(InvalidContentTypeException(contentType, "must not be a zip."))
+  private def contentTypeAnythingBut(contentType: Option[String]): Try[Unit] = {
+    if (contentType.map(_.trim.toLowerCase).exists(str =>
```
rvanheest: would a pattern match make this better readable than an if-else? up to you to decide!

```scala
contentType.map(_.trim.toLowerCase) match {
  case Some(str) if str.nonEmpty
    && !str.startsWith("multipart")
    && !str.matches(contentTypeZipPattern) => Success(())
  case _ => Failure(InvalidContentTypeException(contentType, "must not be application/zip nor start with multipart."))
}
```

PR #141, opened 2019-04-16
EASY-2054 update draft deposit status with value from submitted desposit

jo-pol:
Fixes EASY-2054 update draft deposit status with value from submitted desposit

#### When applied and
* when the status in `drats/USER-ID/UUID/deposit.properties` is SUBMITTED or IN_PROGRESS, it will be updated from the submitted `BAG-STORE.BAG-ID/deposit.properties`
* [x] unit tests dropped in `DepositDirSpec` should be restored an completed in a new `StateManagerSpec` 
* [x] landing page
* [x] easy-home is a better name, like sword does

Read the details of the first commit message.

* [x] Deposit on deasy with deposit-ui and get dataset overview until ingest-flow has completed his task.

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
@@ -52,18 +52,18 @@ case class DepositDir private(draftBase: File, user: String, id: UUID) extends D
    * @param submitBase the base directory with submitted deposits to extract the actual state
    * @return the `StateManager` for this deposit
    */
-  def getStateManager(submitBase: File, landingPageBase: URL): StateManager = {
-    StateManager(bagDir.parent, submitBase, landingPageBase)
+  def getStateManager(submitBase: File, easyHome: URL): StateManager = {
+    StateManager(bagDir.parent, submitBase, easyHome)
   }
 
   /**
    * @param submitBase the base directory with submitted deposits to extract the actual state
    * @return basic information about the deposit.
    */
-  def getDepositInfo(submitBase: File, landingPageBase: URL): Try[DepositInfo] = {
+  def getDepositInfo(submitBase: File, easyHome: URL): Try[DepositInfo] = {
     for {
       title <- getDatasetTitle
-      stateManager = getStateManager(submitBase, landingPageBase)
+      stateManager = getStateManager(submitBase, easyHome)
       stateInfo <- stateManager.getStateInfo
       created = new DateTime(stateManager.draftProps.getString("creation.timestamp")).withZone(UTC)
     } yield DepositInfo(
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
@@ -42,7 +42,7 @@ import scala.util.{ Failure, Success, Try }
  * @param user      the user ID of the deposit's owner
  * @param id        the ID of the deposit
  */
-case class DepositDir private(draftBase: File, user: String, id: UUID, landingPageBase: URL) extends DebugEnhancedLogging {
+case class DepositDir private(draftBase: File, user: String, id: UUID) extends DebugEnhancedLogging {
   val bagDir: File = draftBase / user / id.toString / "bag"
   private val metadataDir = bagDir / "metadata"
   private val depositPropertiesFile = bagDir.parent / "deposit.properties"
@@ -52,7 +52,7 @@ case class DepositDir private(draftBase: File, user: String, id: UUID, landingPa
    * @param submitBase the base directory with submitted deposits to extract the actual state
    * @return the `StateManager` for this deposit
    */
-  def getStateManager(submitBase: File): StateManager = {
+  def getStateManager(submitBase: File, landingPageBase: URL): StateManager = {
     StateManager(bagDir.parent, submitBase, landingPageBase)
   }
 
@@ -60,10 +60,10 @@ case class DepositDir private(draftBase: File, user: String, id: UUID, landingPa
    * @param submitBase the base directory with submitted deposits to extract the actual state
    * @return basic information about the deposit.
    */
-  def getDepositInfo(submitBase: File): Try[DepositInfo] = {
+  def getDepositInfo(submitBase: File, landingPageBase: URL): Try[DepositInfo] = {
     for {
       title <- getDatasetTitle
-      stateManager = getStateManager(submitBase)
+      stateManager = getStateManager(submitBase, landingPageBase)
       stateInfo <- stateManager.getStateInfo
       created = new DateTime(stateManager.draftProps.getString("creation.timestamp")).withZone(UTC)
     } yield DepositInfo(
@@ -171,14 +171,14 @@ object DepositDir {
    * @param user     the user name
    * @return a list of [[DepositDir]] objects
    */
-  def list(draftDir: File, user: String, landingPageBase: URL): Try[Seq[DepositDir]] = {
+  def list(draftDir: File, user: String): Try[Seq[DepositDir]] = {
     val userDir = draftDir / user
     if (userDir.exists)
       userDir
         .list
         .filter(_.isDirectory)
         .map(deposit => Try {
-          DepositDir(draftDir, user, UUID.fromString(deposit.name), landingPageBase)
+          DepositDir(draftDir, user, UUID.fromString(deposit.name))
         }.recoverWith { case t: Throwable => Failure(CorruptDepositException(user, deposit.name, t)) })
         .toSeq
         .collectResults
@@ -193,9 +193,8 @@ object DepositDir {
    * @param id       the identifier of the deposit
    * @return a [[DepositDir]] object
    */
-  def get(draftDir: File, user: String, id: UUID, landingPageBase: URL): Try[DepositDir] = {
-    val depositDir = DepositDir(draftDir, user, id, landingPageBase)
-    println(s"============== ${depositDir.bagDir.parent.exists} ${depositDir.bagDir.parent}")
+  def get(draftDir: File, user: String, id: UUID): Try[DepositDir] = {
+    val depositDir = DepositDir(draftDir, user, id)
     if (depositDir.bagDir.parent.exists) Success(depositDir)
     else Failure(NoSuchDepositException(user, id, new FileNotFoundException()))
   }
@@ -207,9 +206,9 @@ object DepositDir {
    * @param user     the user name
    * @return the newly created [[DepositDir]]
    */
-  def create(draftDir: File, user: String, landingPageBase: URL): Try[DepositDir] = {
+  def create(draftDir: File, user: String): Try[DepositDir] = {
     val depositInfo = DepositInfo()
-    val deposit = DepositDir(draftDir, user, depositInfo.id, landingPageBase)
+    val deposit = DepositDir(draftDir, user, depositInfo.id)
     val depositDir = deposit.draftBase / user / depositInfo.id.toString
     for {
       _ <- Try { depositDir.createDirectories }
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
@@ -24,9 +24,7 @@ import nl.knaw.dans.bag.v0.DansV0Bag
 import nl.knaw.dans.easy.deposit.Errors._
 import nl.knaw.dans.easy.deposit.PidRequesterComponent.{ PidRequester, PidType }
 import nl.knaw.dans.easy.deposit.docs.JsonUtil.toJson
-import nl.knaw.dans.easy.deposit.docs.StateInfo.State
-import nl.knaw.dans.easy.deposit.docs.StateInfo.State.State
-import nl.knaw.dans.easy.deposit.docs.{ StateInfo, _ }
+import nl.knaw.dans.easy.deposit.docs._
 import nl.knaw.dans.lib.error._
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.configuration.PropertiesConfiguration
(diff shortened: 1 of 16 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -89,7 +89,7 @@ class Submitter(stagingBaseDir: File,
       stageDir = (stagingBaseDir / draftDeposit.id.toString).createDirectories()
       stageBag <- DansV0Bag.empty(stageDir / "bag").map(_.withCreated())
       // EASY-1464 3.3.6 change state and copy with the rest of the deposit properties to staged dir
-      submittedId <- draftDeposit.setStateInfo(StateInfo(StateInfo.State.submitted, "Deposit is ready for processing."))
+      submittedId <- stateManager.changeState(StateInfo(StateInfo.State.submitted, "Deposit is ready for processing."))
```
jo-pol: Beyond the scope of this issue. 

We want to copy the new state to the `submitDir` but the new state should be saved in the draft when successfully moved moved from the staged to submitted directory.

src/test/scala/nl.knaw.dans.easy.deposit/StateManagerSpec.scala
```diff
@@ -0,0 +1,134 @@
...
+  it should "result in IllegalStateTransitionException when transitioning from DRAFT to ARCHIVED" in {
+    val props =
+      s"""state.label = DRAFT
+         |state.description = Something's rotten on the state of ...
+         |bag-store.bag-id = $uuid
+         |""".stripMargin
+    draftPropsFile.writeText(props)
+    StateManager(draftDeposit, submitBase)
+      .changeState(StateInfo(State.archived, "rabarbera")) should matchPattern {
+      case e: IllegalStateTransitionException if e.getMessage == "Cannot transition from DRAFT to ARCHIVED" =>
```
jo-pol: @rvanheest Don't see the different between expectations and actual results. Otherwise this pull request is finished.

src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -39,95 +37,31 @@ import scala.util.{ Failure, Success, Try }
...
-  def getDepositInfo: Try[DepositInfo] = {
+  def getDepositInfo(submitBase: File): Try[DepositInfo] = {
     for {
       title <- getDatasetTitle
-      props <- getDepositProps
-      state <- getState(props)
-      created = new DateTime(props.getString("creation.timestamp")).withZone(UTC)
+      stateManager = StateManager(bagDir.parent, submitBase)
+      stateInfo <- stateManager.getStateInfo
+      created = new DateTime(stateManager.draftProps.getString("creation.timestamp")).withZone(UTC)
```
rvanheest: why don't you put this in the `stateManager` as well? Like `stateManager.creationTimestamp`

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  private def saveNewState(newStateInfo: StateInfo) = {
+    draftProps.setProperty(stateLabelKey, newStateInfo.state.toString)
+    draftProps.setProperty(stateDescriptionKey, newStateInfo.stateDescription)
+    draftProps.save()
+    newStateInfo
+  }
+
+  private def getStateLabel(props: PropertiesConfiguration = draftProps): State = {
```
rvanheest: I wonder why you start throwing exceptions here and in the methods below. Why not use proper `Try` wrappers?

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,110 @@
...
+  )
+
+  def getStateInfo: Try[StateInfo] = Try {
+    val draftState = getStateLabel()
+    draftState match {
+      case State.submitted | State.inProgress =>
+        val newState: StateInfo = getProp(stateLabelKey, submittedProps) match {
+          case "SUBMITTED" => StateInfo(State.submitted, getStateDescription())
+          case "REJECTED" => StateInfo(State.rejected, getStateDescription(submittedProps))
+          case "FEDORA_ARCHIVED" => StateInfo(State.archived, "The dataset is published in https://easy.dans.knaw.nl/ui")
```
rvanheest: Where do you get this url from? Not sure if the url is correct or what you try to point to with this description. Besides, should this url be configurable?

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
rvanheest: I don't like the default argument here. By explicitly providing it, it makes you more conscious of where you get the properties from.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,110 @@
...
+    val newState = newStateInfo.state
+    (oldState, newState) match {
+      case (State.draft, State.submitted) |
+           (State.rejected, State.draft) => Success(())
+      case _ => Failure(IllegalStateTransitionException(oldState, newState))
+    }
+  }
+
+  /** @return bag-store.bag-id in case the new state is submitted */
+  def changeState(newStateInfo: StateInfo): Try[UUID] = getStateInfo.flatMap { old =>
```
rvanheest: don't return `Success(null)`, but use `Option`; then the return type would be `Try[Option[UUID]]`

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,110 @@
...
+      case (State.rejected, State.draft) =>
+        draftProps.clearProperty(bagIdKey)
+        saveNewState(newStateInfo)
+        Success(null)
+      case (oldState, newState) =>
+        Failure(IllegalStateTransitionException(oldState, newState))
+    }
+  }
+
+  private def saveNewState(newStateInfo: StateInfo) = {
```
rvanheest: since this just does a side effect, don't return the input parameter

src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
jo-pol: I'm abusing the `stateManager` as it happens to have the properties loaded. Otherwise the `creation.timestamp` has nothing to do with the state.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: It would get complicated to handle the results of `getStateLabel` called outside  and `getStateDescription` called inside the pattern match of `getStateInfo`.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  def getStateInfo: Try[StateInfo] = Try {
+    val draftState = getStateLabel(draftProps)
+    draftState match {
+      case State.submitted | State.inProgress =>
+        val newState: StateInfo = getProp(stateLabelKey, submittedProps) match {
+          case "SUBMITTED" => StateInfo(State.submitted, getStateDescription(draftProps))
+          case "REJECTED" => StateInfo(State.rejected, getStateDescription(submittedProps))
+          case "FEDORA_ARCHIVED" => StateInfo(State.archived, "The dataset is published")
+          case "IN_REVIEW" => StateInfo(State.inProgress, "The dataset is visible for you under your datasets in https://easy.dans.knaw.nl/ui")
```
rvanheest: what happens if we put an `<a>` on this string? does it render as a clickable url in deposit-ui or does it get escaped?

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,117 @@
...
+  )
+
+  def getStateInfo: Try[StateInfo] = Try {
+    val draftState = getStateLabel(draftProps)
+    draftState match {
+      case State.submitted | State.inProgress =>
+        val newState: StateInfo = getProp(stateLabelKey, submittedProps) match {
+          case "SUBMITTED" => StateInfo(State.submitted, getStateDescription(draftProps))
+          case "REJECTED" => StateInfo(State.rejected, getStateDescription(submittedProps))
+          case "FEDORA_ARCHIVED" => StateInfo(State.archived, "The dataset is published")
```
rvanheest: can we communicate the EASY landingpage when the deposit is archived, such that deposit-ui can put that as the anchor on the whole table row?

lindareijnhoudt: you can either point to the DOI, or deposit-api can use the identifier.fedora to assemble the landingspage url

janvanmansum:
Tests EASY-2054

janvanmansum:
EASY-2054 #comment TEst

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -82,36 +93,49 @@ case class StateManager(depositDir: File, submitBase: File) {
...
+   *          IllegalArgumentException (if the value is not a valid UUID)
+   */
   def getSubmittedBagId: Try[UUID] = Try {
     // exception intercepted as not expected
     // by TriedActionResult.getOrRecoverWithActionResult
     UUID.fromString(draftProps.getString(bagIdKey))
   }
 
-  private def saveNewState(newStateInfo: StateInfo) = {
+  // TODO assemble with identifier.fedora
```
rvanheest: Hint: when the dataset is archived, the `state.description` is this url. Perhaps we can make it so that this is also the case for `IN_REVIEW` and `FEDORA_ARCHIVED`? But that needs to happen in `easy-ingest-flow`.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -29,33 +29,41 @@ import scala.util.{ Failure, Success, Try }
...
-    draftState match {
+    getStateLabel(draftProps) match {
       case State.submitted | State.inProgress =>
         val newState: StateInfo = getProp(stateLabelKey, submittedProps) match {
           case "SUBMITTED" => StateInfo(State.submitted, getStateDescription(draftProps))
           case "REJECTED" => StateInfo(State.rejected, getStateDescription(submittedProps))
-          case "FEDORA_ARCHIVED" => StateInfo(State.archived, "The dataset is published")
-          case "IN_REVIEW" => StateInfo(State.inProgress, "The dataset is visible for you under your datasets in https://easy.dans.knaw.nl/ui")
-          case "FAILED" => StateInfo(State.inProgress, "The dataset is in progress")
+          case "FEDORA_ARCHIVED" => StateInfo(State.archived, s"The dataset is published at $landingPage")
```
rvanheest: if we can return the URL as a separate JSON field in these cases that currently have the `$landingPage`, we can render that in an `onClick` on that row. In that case we don't require this `$landingPage` here...

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,147 @@
...
+  def getSubmittedBagId: Try[UUID] = Try {
+    // exception intercepted as not expected
+    // by TriedActionResult.getOrRecoverWithActionResult
+    UUID.fromString(draftProps.getString(bagIdKey))
+  }
+
+  val baseURL = "https://easy.dans.knaw.nl/ui"
+
+  private def landingPage(tryDoi: Boolean) = {
+    Try { getProp("identifier.fedora", submittedProps) }
```
jo-pol: So far never implemented.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: The html-tag is displayed as plain text

rvanheest:
@jo-pol 

> The html-tag is displayed as plain text

There is a way to render this as html, but I'm not yet sure if we should use this method...

https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
rvanheest: There is a way to render this as html, but I'm not yet sure if we should use this method...

https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
rvanheest: To the best of my knowledge, it is `easy-ingest-flow` that sets this property after it received the `easy-dataset:XXX` identifier from Fedora.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: Can't currently check this with a manual test on deasy because ingest fails with
```
ERROR FAILED: unhandled problem during processing. Error: Deposit Agreement Creator Service failed: code=404; msg=DatasetId easy-dataset:17 does not exist
java.lang.RuntimeException: Deposit Agreement Creator Service failed: code=404; msg=DatasetId easy-dataset:17 does not exist
```

src/test/scala/nl.knaw.dans.easy.deposit/StateManagerSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+      s"""state.label = SUBMITTED
+         |state.description = The dataset is ready for processing
+         |bag-store.bag-id = $uuid
+      """.stripMargin)
+    StateManager(draftDeposit, testDir / "does-not-exist").getStateInfo should matchPattern {
+      case Failure(e: PropertyNotFoundException) if e.getMessage ==
+        s"'state.label' not found in $testDir/does-not-exist/$uuid/deposit.properties" =>
+      // actually the submitted deposit does not exist
+      // for example due to lack of space to create a copy in the staging area
+      // or removed by some ignorant clean-up action
```
jo-pol: How to alert the curators/application-manager to check what applies and what to do? For example set the state to rejected to resubmit, change the state to archived or whatever applies.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: That would mean extending the models for StatusInfo and DepositInfo, perhaps another pull request? Merging will need synchronisation with the UI.
https://github.com/DANS-KNAW/easy-deposit-api/blob/5ebe7da4d3c451099c73b2cff1b6e2b13ac97f30/docs/api/api.yml#L723-L755

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: It's dangerous when you put user input in innerHTML. In this case you are not echoing user input. Nevertheless extending the returned json seems a nicer approach IMHO.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: Submitting ate least one payload file works around the problem.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,142 @@
...
+import nl.knaw.dans.easy.deposit.docs.StateInfo
+import nl.knaw.dans.easy.deposit.docs.StateInfo.State
+import nl.knaw.dans.lib.logging.DebugEnhancedLogging
+import org.apache.commons.configuration.PropertiesConfiguration
+
+import scala.util.{ Failure, Success, Try }
+
+case class StateManager(depositDir: File, submitBase: File) extends DebugEnhancedLogging {
+
+  private val landingPageBase = "https://easy.dans.knaw.nl/ui"
```
rvanheest: make this configurable, please
also in `easy-dtap`

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -0,0 +1,141 @@
...
+          case "FAILED" => StateInfo(State.inProgress, s"The deposit is in progress.")
+          case "IN_REVIEW" => StateInfo(State.inProgress, s"The deposit is available at $landingPage")
+          case "FEDORA_ARCHIVED" | "ARCHIVED" => StateInfo(State.archived, s"The dataset is published at $landingPage")
+          case str: String =>
+            logger.error(InvalidPropertyException(stateLabelKey, str, submittedProps).getMessage)
+            StateInfo(State.inProgress, s"The deposit is in progress.")
+        }
+        saveNewState(newState)
+        newState
+      case draftState @ (State.draft | State.rejected | State.archived) =>
```
jo-pol: Once the state became archived, the landing page persists. Even when the service is restarted with another base URL in `application.properties`

PR #146, opened 2019-05-06
EASY-2078: sort deposits on state and date.created

rvanheest:
Fixes EASY-2078

#### When applied it will
* sort deposits on state and date.created

@DANS-KNAW/easy for review

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
@@ -32,6 +32,7 @@ import nl.knaw.dans.easy.deposit.servlets.contentTypeZipPattern
 import nl.knaw.dans.lib.error._
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.configuration.PropertiesConfiguration
+import org.joda.time.DateTime
 import org.scalatra.servlet.MultipartConfig
 
 import scala.util.{ Failure, Success, Try }
@@ -93,7 +94,7 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
   )
 
   // possible trailing slash is dropped
-  private val easyHome: URL = new URL(configuration.properties.getString("easy.home").replaceAll("/?$",""))
+  private val easyHome: URL = new URL(configuration.properties.getString("easy.home").replaceAll("/?$", ""))
 
   @throws[ConfigurationException]("when no existing readable directory is configured")
   private def getConfiguredDirectory(key: String): File = {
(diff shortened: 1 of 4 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -133,11 +134,12 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
    * @return a list of [[docs.DepositInfo]] objects
    */
   def getDeposits(user: String): Try[Seq[DepositInfo]] = {
+    implicit val timestampOrdering: Ordering[DateTime] = Ordering.fromLessThan[DateTime](_ isBefore _).reverse
     for {
       deposits <- DepositDir.list(draftBase, user)
       infos <- deposits.map(_.getDepositInfo(submitBase, easyHome)).collectResults
-    }
-      yield infos
+      sortedInfos = infos.sortBy(deposit => (deposit.state, deposit.date))
```
jo-pol: Please move sorting into `DepositDir.list` and create unit tests.

rvanheest: I can't! `DepositDir.list` only returns `Try[Seq[DepositDir]]`. Only in `DepositDir.getDepositInfo` you get the state from the `StateManager`. It is only after I got the state (in a `DepositInfo` object) that I can sort on `state.label` and `creation.timestamp`.

If you got a better suggestion of where to put this, please let me know!

rvanheest: Thanks for setting up the unit test for this! Really helped with implementing the test itself.

src/test/scala/nl.knaw.dans.easy.deposit/EasyDepositApiAppSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+          // then in-progress deposits
+          deposit5.copy(stateDescription = "The deposit is in progress."),
+          // and finally archived deposits (newest to oldest)
+          deposit6,
+          deposit2,
+        )
+    }
+  }
+
+  private def createDeposit(user: String, title: String, state: State, creationTimestamp: DateTime): DepositInfo = {
```
jo-pol: You are not verifying the user, so it is not a variant for the test, supply a default.
Same for the user. But create an exception to show that deposits of other users are not returned.

```suggestion
  private def createDeposit(user: String = "foo", title: String = "blabla", state: State, creationTimestamp: DateTime): DepositInfo = {
```

src/test/scala/nl.knaw.dans.easy.deposit/EasyDepositApiAppSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+    clearTestDir()
+  }
+
+  "getDeposits" should "list the deposits of a user in sorted order" in {
+    val deposit1 = createDeposit("foo", "my-title1", State.rejected, DateTime.now.minusDays(1).withZone(UTC))
+    val deposit2 = createDeposit("foo", "my-title2", State.archived, DateTime.now.minusDays(2).withZone(UTC))
+    val deposit3 = createDeposit("foo", "my-title3", State.rejected, DateTime.now.minusDays(3).withZone(UTC))
+    val deposit4 = createDeposit("foo", "my-title4", State.draft, DateTime.now.minusDays(4).withZone(UTC))
+    val deposit5 = createDeposit("foo", "my-title5", State.inProgress, DateTime.now.minusDays(5).withZone(UTC))
+    val deposit6 = createDeposit("foo", "my-title6", State.archived, DateTime.now.minusDays(6).withZone(UTC))
```
jo-pol: Don't repeat yourself with the dates. The method `createDeposit` can turn an int into a date.

PR #148, opened 2019-05-13
EASY-2109 skip corrupt deposits

jo-pol:
Fixes EASY-2109 skip corrupt deposits

#### When applied it will
* Won't fail when one of the deposit directories is corrupt
* 
* 

* add a directory between the drafts of a user that is not a UUID,
* try to show the list of deposits
* it will no longer fail, the problematic directory is logged

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
@@ -176,10 +176,11 @@ object DepositDir {
     if (userDir.exists)
       userDir
         .list
-        .filter(_.isDirectory)
+        .withFilter(_.isDirectory)
         .map(deposit => Try {
           DepositDir(draftDir, user, UUID.fromString(deposit.name))
         }.recoverWith { case t: Throwable => Failure(CorruptDepositException(user, deposit.name, t)) })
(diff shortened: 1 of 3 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/DepositDir.scala
```diff
@@ -176,10 +176,11 @@ object DepositDir {
     if (userDir.exists)
       userDir
         .list
-        .filter(_.isDirectory)
+        .withFilter(_.isDirectory)
         .map(deposit => Try {
           DepositDir(draftDir, user, UUID.fromString(deposit.name))
         }.recoverWith { case t: Throwable => Failure(CorruptDepositException(user, deposit.name, t)) })
+        .withFilter(_.isSuccess)// CorruptDepositException logged itself
```
janvanmansum: What other type of Failures may result from this? You are doing a `collectResults` at the end, so I gather there may still be Failures in the sequence (otherwise the `collectResults` would not really make sense).

janvanmansum: Come to think of it, you are filtering all types of Failures, so the `collectResults` will never collect any Failures.

PR #157, opened 2019-06-06
EASY-2092: POSTs MAY overwrite existing files

lindareijnhoudt:
Fixes EASY-2029

#### When applied it will
* Remove 'AlreadySubmitted' error message on POST/PUT
* Allow POST to overwrite files

### DIFF
--- a/docs/api/api.yml
+++ b/docs/api/api.yml
@@ -329,8 +329,6 @@ paths:
           $ref: "#/components/responses/ForbiddenUpdate"
         404:
           $ref: "#/components/responses/NotFound"
-        409:
-          $ref: "#/components/responses/AlreadySubmitted"
         500:
           $ref: "#/components/responses/InternalServerError"
       security:
@@ -396,7 +394,7 @@ paths:
     post:
       tags:
       - directory
-      summary: Adds one or more files.
+      summary: Adds or overwrites one or more files.
       description: |
         The `Content-Type` must always be `multipart/form-data`, which means the body is formatted using the
         [multipart media type](https://tools.ietf.org/html/rfc2046.html#section-5.1); each part contains one file.
@@ -407,8 +405,8 @@ paths:
         unzipped in `dir_path`. For each file in the ZIP file, the path in the deposit will be `dir_path`/`path_in_zip`.
         The ZIP file may contain `nested ZIP files`, which will **not** be extracted, but stored as ZIP file.
 
-        Concurrent POST calls to the same deposit are not allowed. The result of concurrent PUT and POST calls is
-        undefined.
+        Existing files at the same location will be overwritten. Concurrent POST or PUT calls to the same deposit are not allowed.
+
       parameters:
       - $ref: "#/components/parameters/DepositId"
       - $ref: "#/components/parameters/DirPath"
@@ -437,9 +435,7 @@ paths:
           $ref: "#/components/responses/Unauthorized"
         404:
           $ref: "#/components/responses/NotFound"
-        409.1:
-          $ref: "#/components/responses/ConflictOverwritesFilesOnServer"
-        409.2:
+        409:
           $ref: "#/components/responses/ConflictConcurrentPostNotAllowed"
         500:
           $ref: "#/components/responses/InternalServerError"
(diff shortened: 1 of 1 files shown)


docs/api/api.yml
```diff
@@ -329,8 +329,6 @@ paths:
           $ref: "#/components/responses/ForbiddenUpdate"
         404:
           $ref: "#/components/responses/NotFound"
-        409:
-          $ref: "#/components/responses/AlreadySubmitted"
```
jo-pol: I guess you dropped it in favour of `ForbiddenStateTransition`

docs/api/api.yml
```diff
@@ -533,8 +530,10 @@ paths:
           $ref: "#/components/responses/ForbiddenUpdate"
         404:
           $ref: "#/components/responses/NotFound"
-        409:
-          $ref: "#components/responses/ConflictAttemptToOverwriteDirectoryWithFile"
+        409.1:
+          $ref: "#/components/responses/ConflictConcurrentPostNotAllowed"
```
jo-pol: Concurrent uploads would cover both types of requests

docs/api/api.yml
```diff
@@ -642,10 +641,6 @@ components:
       description: |
         Forbidden. Illegal state transition. (Only allowed to transition from DRAFT to SUBMITTED and from REJECTED to DRAFT.)
 
-    AlreadySubmitted:
-      description: |
-        Conflict. The deposit already exists in the submit area. Possibly due to a resubmit.
```
jo-pol: Dropped in favour of `ForbiddenUpdate`?

docs/api/api.yml
```diff
@@ -656,7 +651,7 @@ components:
       description: Conflict. The following file(s) already exist on the server
 
     ConflictConcurrentPostNotAllowed:
-      description: Conflict. Concurrent POSTs to the same deposit are not allowed.
+      description: Conflict. Concurrent POSTs and PUTs to the same deposit are not allowed.
```
jo-pol:
```suggestion
      description: Conflict. Concurrent POSTs and PUTs of file uploads to the same deposit are not allowed.
```

docs/api/api.yml
lindareijnhoudt:
```suggestion
          $ref: "#/components/responses/ConflictConcurrentUploadsNotAllowed"
```

docs/api/api.yml
lindareijnhoudt:
```suggestion
      description: Conflict. Concurrent file uploads to the same deposit are not allowed.
```

docs/api/api.yml
lindareijnhoudt: You like this distinction between AlreadySubmitted and ForbiddenUpdate? If we are explicit in the error message, i think they are conveying the same message

docs/api/api.yml
```diff
@@ -656,7 +651,7 @@ components:
       description: Conflict. The following file(s) already exist on the server
 
     ConflictConcurrentPostNotAllowed:
```
lindareijnhoudt:
```suggestion
    ConflictConcurrentUploadsNotAllowed":
```

docs/api/api.yml
```diff
@@ -437,9 +435,7 @@ paths:
           $ref: "#/components/responses/Unauthorized"
         404:
           $ref: "#/components/responses/NotFound"
-        409.1:
-          $ref: "#/components/responses/ConflictOverwritesFilesOnServer"
-        409.2:
+        409:
           $ref: "#/components/responses/ConflictConcurrentPostNotAllowed"
```
lindareijnhoudt:
```suggestion
          $ref: "#/components/responses/ConflictConcurrentUploadsNotAllowed"
```

docs/api/api.yml
```diff
@@ -407,8 +405,8 @@ paths:
         unzipped in `dir_path`. For each file in the ZIP file, the path in the deposit will be `dir_path`/`path_in_zip`.
         The ZIP file may contain `nested ZIP files`, which will **not** be extracted, but stored as ZIP file.
 
-        Concurrent POST calls to the same deposit are not allowed. The result of concurrent PUT and POST calls is
-        undefined.
+        Existing files at the same location will be overwritten. Concurrent POST or PUT calls to the same deposit are not allowed.
```
janvanmansum:
```suggestion
        Existing files at the target location will be overwritten. Concurrent POST or PUT calls to the same deposit are not allowed.
```

docs/api/api.yml
jo-pol: I just was surprised as it seemed off topic. But you are right about conveying the same message. Just wanted to be sure

PR #159, opened 2019-06-14
EASY-2152 : upgrade jetty with fix for cleanup of multipart on aborted upload

jo-pol:
https://github.com/eclipse/jetty.project/issues/3683
https://github.com/eclipse/jetty.project/releases/tag/jetty-9.4.19.v20190610

Fixes EASY-2152 : upgrade jetty with fix for cleanup of multipart on aborted upload

#### When applied it will
* remove multipart files of aborted uploads
* 
* 

on deasy
* `watch ls -la /var/opt/dans.knaw.nl/tmp/easy-deposit-api/multipart-location/`
* upload a large file
* abort before the status bar reaches 100%
* before deploy the multipart file is not removed, after deploy it will be removed

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,5 +8,3 @@ Welcome
 -------
 
 Welcome to the `easy-deposit-api` project. See the [GitHub pages](https://dans-knaw.github.io/easy-deposit-api/) for the manual and API documentation.
-
-
--- a/pom.xml
+++ b/pom.xml
@@ -79,10 +79,12 @@
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-server</artifactId>
+            <version>9.4.19.v20190610</version>
         </dependency>
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-servlet</artifactId>
+            <version>9.4.19.v20190610</version>
         </dependency>
         <dependency>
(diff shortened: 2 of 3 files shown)


rvanheest:
I've added Jetty to our dependency monitoring in Slack. Now, once a new version is released, we get an automatic notification about this.

rvanheest:
@jo-pol not sure, but would it be possible to write a test that fails on the old dependency version but succeeds on this one? (don't spend to much time on it, though, if you can't immediately think of a nice solution!)

jo-pol:
I'm afraid I have no inspiration for a unit test. 

The test would have to
* configure a `multipart.file-size-threshold` of zero
* monitor the files in `multipart.location`
* before a simulated interrupt some but not all bytes of the bodypart should have been saved to a file in the `multipart.location`
* another test should have multiple body parts and get interrupted somewhere after the first and before the last part was saved.

I also wonder it would test real jetty behaviour.

https://github.com/DANS-KNAW/easy-deposit-api/blob/dd284b99886f8d34fdc6717a937888b25ea8a930/src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala#L170-L185

jo-pol:
Writing the phrase "I also wonder it would test real jetty behaviour": we should only write unit tests to check our own code, not test jetty-code. So even if I could, I should not.

PR #162, opened 2019-07-01
EASY-2187 worker thread for submits and SHA calculations

jo-pol:
Fixes EASY-2158 worker thread set up

#### To do in/with the WorkerThread class
* [x] Make `val stagedBaseDir` configurable for unit tests and implement the stub `getDepositState`.
* [ ] `calculateShas` and `finalizeSubmit` are still stubs, logging will show what the worker thread would do.
* [ ] Allow just a single thread for now: all requests for one deposit should be executed by the same thread, otherwise SHA-calulations and submits might get executed in parallel.
* "[ordering is highly implementation specific](https://docs.oracle.com/javase/7/docs/api/java/nio/file/WatchService.html)": so SHA calculations might come after submission.
* "Short-lived files may not be detected" so SHA's might not get calculated for uploads. SHA calculations interrupted by a hard shutdown get lost anyway.

The new class `WorkerThread`.

Running a test

* Make comment of the override with `mockedLogger` in `WorkerThreadSpec`.
* Once `WorkerThread` is started/stopped by `EasyDepositApiService` you have two methods to run locally:
   * build and execute `dans-dev-tools/run-service.sh`
  * VM options in Intellij
    ```
    -Xmx1g -Dapp.home=home -Dlogback.configurationFile=home/cfg/logback.xml
    ```
  run-config

Scenarios:

* Use normal submits/uploads, when running the service locally you have no ui, use something like:
  ```
  curl -i  -u user001:user001 -X ... "http://localhost:20190/deposit/<UUID>/..."
  ```
* kill-restart the service while uploads/submit are still in progress by the `WorkerThread`. Depending on how the service was started, you can get the process number to kill with
  `service easy-deposit-api status` or `ps -A|grep deposit-api`


src/main/assembly/dist/cfg/application.properties
```diff
@@ -19,24 +19,29 @@ multipart.file-size-threshold=3145728
 #
 # Directories for managing the deposits.
 #
-# - stage-zips:       is used for uploads of zipped directories or complete deposits.
+# - staged:           Directories have two functions:
```
janvanmansum:
```suggestion
# - staged:           Sub-directories have two functions:
```

src/main/assembly/dist/cfg/application.properties
```diff
@@ -19,24 +19,29 @@ multipart.file-size-threshold=3145728
...
 #
-# N.B. multipart.location, stage-zips and drafts must all be located on the same partition
+# N.B. multipart.location, staged, drafts and submit-to must all be located on the same partition
 # in order to support atomic moving.
-# The same holds for the couple stage-for-submit and submit-to.
 #
-# Note that the stage-* directories only have temporary content, so the service should clear them
-# automatically on completion of a request.
+# Note that a directory in the staged folder only has temporary content, so the service should
+# clear it automatically on completion of a request. The error message of a startup failure explains
```
janvanmansum:
```suggestion
# remove it automatically on completion of a request. The error message of a startup failure explains
```

src/main/scala/nl.knaw.dans.easy.deposit/EasyDepositApiApp.scala
```diff
@@ -64,25 +64,20 @@ class EasyDepositApiApp(configuration: Configuration) extends DebugEnhancedLoggi
...
-    if (dir.nonEmpty) {
-      // TODO move to Validation and/or new Uploader class for DepositServlet post("/:uuid/file/*")
-      throw new FileAlreadyExistsException(
-        s"Upload staging area [$dir] should be empty unless force shutdown during an upload request. Check logging related to 'POST /deposit/{id}/file/{dir_path}'."
-      )
-    }
+  private val stagedDir = {
+    val dir = getConfiguredDirectory("deposits.staged")
+    if (dir.nonEmpty) throw LeftoversOfForcedShutdownException(dir)
+    logger.info(s"Uploads/submits are staged in $dir")
```
janvanmansum:
```suggestion
    logger.info(s"Uploads/submits will be staged in $dir")
```

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
```diff
@@ -0,0 +1,129 @@
...
+  val stagedBaseDir = File("target")
+
+  def main(args: Array[String]): Unit = {
+
+    // TODO create watchService as first action in EasyDepositApiService.start
+    //  close it as last action in EasyDepositApiService.stop
+    stagedBaseDir.watchService.apply { watchService =>
+      stagedBaseDir.register(watchService, Seq(ENTRY_CREATE, ENTRY_DELETE))
+      handleAbortedByHardShutdown() // between register and take of watchService!
+      while ( { // TODO in a sub-thread (possibly with handleAbortedByHardShutdown, for a quick start)
```
janvanmansum: Implement with `Iterator.continually()`?

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
```diff
@@ -0,0 +1,129 @@
...
+        // https://howtodoinjava.com/java8/java-8-watchservice-api-tutorial/
+        // take, because polling skips staged-upload-dirs created and deleted within a single cycle
+        Try(watchService.take())
+          .map(handleEvents)
+          .doIfFailure { case e =>
+            // does not get here on SIGINT despite InterruptedException thrown by take
```
janvanmansum: Isn't this to be expected? `Try.map` does not catch exceptions to turn them into `Failure`s.

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
```diff
@@ -0,0 +1,129 @@
...
+      getDepositState(abortedDir.name) match {
+        case State.draft => abortedUpload(abortedDir)
+        case State.finalizing => abortedSubmit(abortedDir)
+        case state => logger.error(s"Staged garbage : $state $abortedDir")
+      }
+    }
+    logger.info(s"So far what was found at startup in $stagedBaseDir")
+  }
+
+  private def abortedSubmit(stagedSubmitDir: File): Unit = {
```
janvanmansum:
```suggestion
  private def retrySubmit(stagedSubmitDir: File): Unit = {
```

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  private def handleEvents(watchKey: WatchKey): Boolean = {
+    watchKey.pollEvents.stream.forEach(handleEvent(_))
+    watchKey.reset() // next pending event
+  }
+
+  private def handleEvent(event: WatchEvent[_]): Unit = {
+    val stagedDir = stagedBaseDir / event.context().toString
```
janvanmansum:
```suggestion
    val stagedDir = stagedBaseDir / event.context.toString
```

unless `context()` has side effects.

janvanmansum:
```suggestion
    val stagedDepositDir = stagedBaseDir / event.context().toString
```

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
```diff
@@ -0,0 +1,129 @@
...
+
+  def deleteContent(dir: File): Unit = {
+    dir.list.foreach { f =>
+      println(s"deleting $f")
+      if (f.isDirectory) deleteContent(f)
+      f.delete()
+    }
+  }
+
+  private def abortedUpload(stagedUploadDir: File): Unit = {
```
janvanmansum: Can you now simple recursively remove `stagedUploadDir`?

jo-pol: I don't know of a delete method with such an option.

For submits I should keep the directory but drop its content. For uploads I need to drop the directory with its content. Other content than multipart files might indicate some problem, so I left it this way once I implemented deleteContent.

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
jo-pol: Both `aborted` methods start with logging `found ... aborted by hard shutdown`. They are called side-by-side from `handleAbortedByHardShutdown`. So I choose to call both methods `aborted`. But indeed one retries it, the other only triggers a retry to be executed via the `watchService.take` loop.

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
jo-pol: The `Try.map` returns a `Try[boolean]` so `doIfFailure` can still process exceptions thrown by `take`. I suppose `SIGINT` is some hard shutdown (killing a process) of this class. I don't know yet how to trigger the InterruptException for the watchService. It's new for me too.

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
jo-pol: Looked at https://stackoverflow.com/questions/13504461/repeating-function-call-until-well-get-non-empty-option-result-in-scala
Tried `Stream.continually{...}.dropWhile(_)` but `...` wasn't executed even once.

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
janvanmansum: My mistake, I misread your comment. I though you were talking about exceptions thrown in `handleEvents`.

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
```diff
@@ -0,0 +1,136 @@
...
+      handleWatchService(watchService) // TODO in a sub-thread created by EasyDepositApiService.start
+    }
+  }
+
+  def handleWatchService(watchService: WatchService): Unit = {
+    // https://howtodoinjava.com/java8/java-8-watchservice-api-tutorial/
+    // take, because polling skips staged-upload-dirs created and deleted within a single cycle
+    handleAbortedByHardShutdown()
+    Iterator.continually(()).foreach { _ =>
+      Try(watchService.take())
```
jo-pol: @rvanheest did you try `take` in the past or `poll`? If the first I hope you remember the problems. Note also my comments in `handleWatchKey`, I don't understand every detail there.

rvanheest: I remember using the `watchService` indirectly in combination with RxJava. See their [implementation](https://github.com/ReactiveX/RxJavaFileUtils/blob/3ee8d12fc2947eb5d7e65cfb13bfd6ea13be884e/src/main/java/rx/fileutils/FileSystemWatcher.java#L72-L77) for some inspiration. I think you should use `register` for 'push-based' behavior, rather than pulling continuously...

rvanheest: Looking somewhat further, though, I see they use `take` as well... https://github.com/ReactiveX/RxJavaFileUtils/blob/3ee8d12fc2947eb5d7e65cfb13bfd6ea13be884e/src/main/java/rx/fileutils/FileSystemWatcher.java#L102-L124

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
```diff
@@ -0,0 +1,151 @@
...
+
+import scala.util.Try
+
+object WorkerThread extends DebugEnhancedLogging {
+
+  val stagedBaseDir = File("target") // TODO allow inject/config to unit-test takeLoop
+
+  def main(args: Array[String]): Unit = {
+
+    // TODO create watchService as first action in ServiceStarter.start
```
jo-pol:
```suggestion
    // TODO create watchService as first action in EasyDepositApiService.start
```

src/main/scala/nl.knaw.dans.easy.deposit/WorkerThread.scala
janvanmansum: Doesn't have`better.files` such an option? If not, maybe just use commons-io: https://commons.apache.org/proper/commons-io/javadocs/api-release/index.html

janvanmansum:
I am closing this one, as it has been stale for too long. We can still reference it for the main ideas if we want to, once we get around implementing this issue.

PR #167, opened 2019-08-02
EASY-2222 overide stacktrace by ingest-flow with mail-to-dans message

jo-pol:
Fixes EASY-2222 overide stacktrace by ingest-flow with mail-to-dans message

#### When applied it will
* generate mailto-links with two variants of generated references:
  * one with the UUID found with `bag-store.bag-id` in the `deposit.properties of the draft deposit`
  * one with 'DRAFT/<user>/<UUID>' as fallback if the property (or submitted copy of the draft) was not found
* [x] fill in the title and test a long one

The template in `StateManager.mailtoDansMssage` and when the method is called 

* [x] on deasy: 
  * `easy-dtap` and `easy-app` on ui-integration branch
  * build/deploy `easy-deposit-api` and `easy-webui`
  * shut down a service like `easy-validate-dans-bag `
  * submit a deposit (with at least one file uploaded)
  * refresh the list of deposits until ingest-flow is done
* Try the `expectedURL`-s in `StateManagerSpec`, result is shown in the screenshots. 

#### Screen shots
* The link only shows up in the notes of the deposit-ui when the mouse hovers over it
* Note that the deposit-ui does not show the user-id. The user-id does appear in the the message greeting. Users might change it and/or an automated signature gets added.
* Compare the trimmed titles in the `dataset.json` snippets with the subject line and message body.
image
image

https://github.com/DANS-KNAW/easy-deposit-api/blob/9ee217567ee89d979bd26c977684637d94276e7c/src/test/scala/nl.knaw.dans.easy.deposit/StateManagerSpec.scala#L132-L133
image

https://github.com/DANS-KNAW/easy-deposit-api/blob/9ee217567ee89d979bd26c977684637d94276e7c/src/test/scala/nl.knaw.dans.easy.deposit/StateManagerSpec.scala#L150-L156
image

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
@@ -24,11 +24,14 @@ import nl.knaw.dans.easy.deposit.docs.StateInfo
 import nl.knaw.dans.easy.deposit.docs.StateInfo.State
 import nl.knaw.dans.lib.logging.DebugEnhancedLogging
 import org.apache.commons.configuration.PropertiesConfiguration
+import org.scalatra.util.UrlCodingUtils.queryPartEncode
 
 import scala.util.{ Failure, Success, Try }
 
-case class StateManager(depositDir: File, submitBase: File, easyHome: URL) extends DebugEnhancedLogging {
+case class StateManager(draftDeposit: DepositDir, submitBase: File, easyHome: URL) extends DebugEnhancedLogging {
 
+  private val depositDir = draftDeposit.bagDir.parent
+  private val relativeDraftDir = s"${ draftDeposit.user }/${ draftDeposit.id }"
   private val stateDescriptionKey = "state.description"
   private val stateLabelKey = "state.label"
 
@@ -39,9 +42,10 @@ case class StateManager(depositDir: File, submitBase: File, easyHome: URL) exten
   val draftProps = new PropertiesConfiguration(
     (depositDir / "deposit.properties").toJava
   )
-  private lazy val submittedProps = new PropertiesConfiguration(
-    (submitBase / getProp(bagIdKey, draftProps) / "deposit.properties").toJava
-  )
+  private lazy val submittedProps = Try(getProp(bagIdKey, draftProps))
+    .map(submittedId =>
+      new PropertiesConfiguration((submitBase / submittedId.toString / "deposit.properties").toJava)
+    ).getOrElse(new PropertiesConfiguration)
 
   /** @return the state-label/description from drafts/USER/UUID/deposit.properties
    *          unless more recent values might be available in SUBMITTED/UUID/deposit.properties
@@ -51,17 +55,21 @@ case class StateManager(depositDir: File, submitBase: File, easyHome: URL) exten
       case stateInDraftDeposit @ (State.submitted | State.inProgress) =>
         val stateInSubmittedDeposit: StateInfo = Try(getProp(stateLabelKey, submittedProps)) match {
           case Success("SUBMITTED") => StateInfo(State.submitted, getStateDescription(draftProps))
-          case Success("REJECTED") => StateInfo(State.rejected, getStateDescription(submittedProps))
-          case Success("FAILED") => StateInfo(State.inProgress, s"The deposit is in progress.")
-          case Success("IN_REVIEW") => StateInfo(State.inProgress, s"""The deposit is available at <a href="$landingPage" target="_blank">$landingPage</a>""")
+          case Success("REJECTED") => Try { getProp("curation.performed", submittedProps) } match {
+            case Success("yes") => StateInfo(State.inProgress, getStateDescription(submittedProps))
+            case _ => StateInfo(State.inProgress, mailToDans)
+          }
+          case Success("FAILED") => StateInfo(State.inProgress, mailToDans)
+          case Success("IN_REVIEW") => StateInfo(State.inProgress, landingPage("deposit is available"))
           case Success("FEDORA_ARCHIVED") |
-               Success("ARCHIVED") => StateInfo(State.archived, s"""The dataset is published at <a href="$landingPage" target="_blank">$landingPage</a>""")
+               Success("ARCHIVED") => StateInfo(State.archived, landingPage("dataset is published"))
           case Success(str: String) =>
             logger.error(InvalidPropertyException(stateLabelKey, str, submittedProps).getMessage)
-            StateInfo(State.inProgress, s"The deposit is in progress.")
+            StateInfo(State.inProgress, mailToDans)
           case Failure(e) =>
-            logger.error(e.getMessage, e)
-            StateInfo(stateInDraftDeposit, getStateDescription(draftProps))
+            logger.error(s"Could not find state of submitted deposit [draft = $relativeDraftDir]: ${e.getMessage}")
+            // return as we don't want to change anything anymore when we are in such deep trouble
+            return Success(StateInfo(stateInDraftDeposit, mailToDans))
         }
         saveNewStateInDraftDeposit(stateInSubmittedDeposit)
         stateInSubmittedDeposit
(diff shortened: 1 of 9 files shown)


src/test/scala/nl.knaw.dans.easy.deposit/StateManagerSpec.scala
```diff
@@ -104,11 +109,87 @@ class StateManagerSpec extends TestSupportFixture {
...
+         |bag-store.bag-id = $submittedUuid
+      """.stripMargin)
+    submittedPropsFile.writeText(
+      s"""state.label = REJECTED
+         |curation.performed = no
+      """.stripMargin)
+    ((draftDeposit.bagDir / "metadata").createDirectories() / "dataset.json")
+      .write(
+        """{"titles":["
+          | A complete ground hog <a href='mailto:info@dans.knaw.nl'>weather</a> report
```
jo-pol: The `easy-deposit-ui` happily writes HTML-tags in the `dataset.json`
It doesn't hurt plain text messages but I don't feel at ease.
Also: what happens when transformed to DDM and further down the line?

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -23,51 +23,56 @@ import nl.knaw.dans.easy.deposit.Errors.{ IllegalStateTransitionException, Inval
...
+        case Success("REJECTED") => getProp("curation.performed", submittedProps) match {
+          case Success("yes") => saveNewStateInDraftDeposit(StateInfo(State.inProgress, getStateDescription(submittedProps)))
+          case _ => StateInfo(stateInDraftDeposit, mailToDansMessage)
         }
-        saveNewStateInDraftDeposit(stateInSubmittedDeposit)
-        stateInSubmittedDeposit
-      case draftState @ (State.draft | State.rejected | State.archived) =>
-        StateInfo(draftState, getStateDescription(draftProps))
-    }
+        case Success("FAILED") => StateInfo(State.inProgress, mailToDansMessage)
```
jo-pol:
```suggestion
        case Success("FAILED") => StateInfo(stateInDraftDeposit, mailToDansMessage)
```

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -115,31 +119,57 @@ case class StateManager(depositDir: File, submitBase: File, easyHome: URL) exten
...
+    val body = queryPartEncode(
+      s"""Hello
+         |
+         |Could you please figure out what went wrong with my deposit?
+         |
+         |It has reference:
+         |   $ref
+         |and title:
+         |   $mediumTitle
+         |""".stripMargin
```
rvanheest:
```suggestion
         |
         |Kind regards,
         |$userFullName""".stripMargin
```

rvanheest: the `userFullName` is not yet defined; I'm not sure if it is easily doable to fetch the user's full name here. Otherwise maybe the username will do?

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -115,31 +119,57 @@ case class StateManager(depositDir: File, submitBase: File, easyHome: URL) exten
...
+    val subject = queryPartEncode(s"${ shortTitle } (reference nr: $ref)")
+    val body = queryPartEncode(
+      s"""Hello
+         |
+         |Could you please figure out what went wrong with my deposit?
+         |
+         |It has reference:
+         |   $ref
+         |and title:
+         |   $mediumTitle
```
rvanheest: perhaps also add the username here, if easily obtainable.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -23,51 +23,55 @@ import nl.knaw.dans.easy.deposit.Errors.{ IllegalStateTransitionException, Inval
...
-            logger.error(InvalidPropertyException(stateLabelKey, str, submittedProps).getMessage)
-            StateInfo(State.inProgress, s"The deposit is in progress.")
-          case Failure(e) =>
-            logger.error(e.getMessage, e)
-            StateInfo(stateInDraftDeposit, getStateDescription(draftProps))
+  def getStateInfo: Try[StateInfo] = getStateInDraftDeposit.map {
+    case draftState @ (State.draft | State.rejected | State.archived) =>
+      StateInfo(draftState, getStateDescription(draftProps))
+    case stateInDraftDeposit @ (State.submitted | State.inProgress) =>
+      getProp(stateLabelKey, submittedProps) match {
```
rvanheest: maybe outside the scope of this issue, but a `.map` with pattern match would be less repetitious that this match with `case Success(...)`

rvanheest: Also a match on enum values would maybe be more desirable than matching on strings?

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -115,31 +119,57 @@ case class StateManager(depositDir: File, submitBase: File, easyHome: URL) exten
...
+         |
+         |Could you please figure out what went wrong with my deposit?
+         |
+         |It has reference:
+         |   $ref
+         |and title:
+         |   $mediumTitle
+         |""".stripMargin
+    )
+    s"""Something went wrong while processing this deposit. Please <a href="mailto:info@dans.knaw.nl?subject=$subject&body=$body">contact DANS</a>"""
```
rvanheest: Not sure if my concern makes any sense, but should we keep this address hardcoded? (1) scrapers could find this address and start sending spam to it, (2) if we want to change the address later, it is easier to do so if it is configurable.

src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala
```diff
@@ -349,7 +349,8 @@ class UploadSpec extends DepositServletFixture {
     ) {
       absoluteTarget.list.size shouldBe 2
       status shouldBe CONFLICT_409
-      body shouldBe s"The following file(s) already exist on the server: some/3.txt, some/2.txt"
+      val prefix = "The following file(s) already exist on the server:"
+      body should (equal(s"$prefix some/3.txt, some/2.txt".toString) or equal(s"$prefix some/2.txt, some/3.txt".toString))
```
rvanheest: isn't this a string already? Why use `.toString`?

src/test/scala/nl.knaw.dans.easy.deposit/StateManagerSpec.scala
rvanheest: Good point! Please make an issue for that or ask @rkruithof if he has already tested that...

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: I'm afraid only ldap can provide the full user name. Indeed quite elaborate and probably error-prone.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: Ideally the enum should somehow be shared with ingest-flow. I doubt we have a good mechanism for that purpose.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
rvanheest: I was afraid you would say that indeed 😉 
Perhaps the username would suffice?

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
rvanheest: Not sure if it's a good idea to have a shared enum (a.k.a. a library) for this. Hopefully in the near future we will move this to `easy-deposit-properties`, whose GraphQL interface **guarantees** a specific set of values.

src/test/scala/nl.knaw.dans.easy.deposit/servlets/UploadSpec.scala
jo-pol: Some subtle difference made the test fail.

src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
jo-pol: Looked into that. Found only existing hard-coded examples in our code, but not sure my search-result was complete.

src/test/scala/nl.knaw.dans.easy.deposit/StateManagerSpec.scala
jo-pol: An altered test shows that DDM encodes the `<` as `&lt;`. So that part is safe.

jo-pol: Now further trimming a title as of the first `<` within the 42 characters.

PR #170, opened 2019-08-12
EASY-2205: Map alternative identifiers to isFormatOf

lindareijnhoudt:
Fixes EASY-2205

#### When applied it will
* Map alternative identifiers to isFormatOf instead of identifier
* Use non-versioned ddm.xsd
* Add testcases

### DIFF
--- a/src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
+++ b/src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
@@ -107,18 +107,14 @@ object DDM extends SchemedXml with DebugEnhancedLogging {
       case Relation(_, None, Some(title: String)) => <label xml:lang={ lang }>{ title }</label>
       case RelatedIdentifier(Some("id-type:DOI"), Some(value), _) => <label scheme="id-type:DOI" href={ "https://doi.org/" + value }>{ value }</label>
       case RelatedIdentifier(Some("id-type:URN"), Some(value), _) => <label scheme="id-type:URN" href={ "http://persistent-identifier.nl/" + value }>{ value }</label>
-      case RelatedIdentifier(Some("id-type:URI"), Some(value), _) => <label scheme="id-type:URI" href={ value }>{ value }</label>
-      case RelatedIdentifier(Some("id-type:URL"), Some(value), _) => <label scheme="id-type:URL" href={ value }>{ value }</label>
+      case RelatedIdentifier(Some(scheme @ ("id-type:URI" | "id-type:URL")), Some(value), _) => <label scheme={ scheme } href={ value }>{ value }</label>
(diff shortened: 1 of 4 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
```diff
@@ -105,12 +105,16 @@ object DDM extends SchemedXml with DebugEnhancedLogging {
       case Relation(_, Some(url: String), Some(title: String)) => <label xml:lang={ lang } href={ url }>{ title }</label>
       case Relation(_, Some(url: String), None) => <label href={ url }>{ url }</label>
       case Relation(_, None, Some(title: String)) => <label xml:lang={ lang }>{ title }</label>
-      case RelatedIdentifier(scheme, Some(value), _) => <label xsi:type={ scheme.nonBlankOrNull }>{ value }</label>
+      case RelatedIdentifier(Some("id-type:DOI"), Some(value), _) => <label scheme="id-type:DOI" href={ "https://doi.org/" + value }>{ value }</label>
+      case RelatedIdentifier(Some("id-type:URN"), Some(value), _) => <label scheme="id-type:URN" href={ "http://persistent-identifier.nl/" + value }>{ value }</label>
+      case RelatedIdentifier(Some(scheme @ ("id-type:URI" | "id-type:URL")), Some(value), _) => <label scheme={ scheme } href={ value }>{ value }</label>
+      case RelatedIdentifier(scheme, Some(value: String), _) => <label xsi:type={ scheme.nonBlankOrNull }>{ value }</label>
```
rvanheest: Is `: String` really necessary. Otherwise this might be a partial pattern match...

janvanmansum: It is not necessary, but it is also not a partial pattern match, because there is only one constructor and this parameter is an `Option[String]`. Anyway, let's remove it, as it seems to be more idiomatic to leave type specs out in these cases.

src/test/resources/manual-test/datasetmetadata-from-ui-all.json
```diff
@@ -127,6 +127,10 @@
     {
       "scheme": "id-type:ARCHIS-ZAAK-IDENTIFICATIE",
       "value": "archis nr. 2"
+    },
+    {
+      "scheme": "id-type:DOI",
+      "value": "10.5072/test-doi"
```
janvanmansum: Schouldn't we get rid of the DOI test-namespace? Just to be sure...

src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
rvanheest:
```suggestion
      case RelatedIdentifier(scheme, Some(value), _) => <label xsi:type={ scheme.nonBlankOrNull }>{ value }</label>
```

PR #173, opened 2019-08-19
EASY-2261 zip with empty dir

jo-pol:
Fixes EASY-2261 zip with empty dir

#### When applied it will
* recursively remove empty directories (as result of a zipped upload) from the staging directory
* have some assertions moved from servlet tests to `RichZipInputStreamSpec.unzipPlainEntriesTo` 
* 


src/main/scala/nl.knaw.dans.easy.deposit/servlets/package.scala
```diff
@@ -84,6 +81,19 @@ package object servlets extends DebugEnhancedLogging {
         }
       }
 
+      @tailrec
+      def cleanup(file: File): Unit = {
+        // in case walk returns a parent after its children
+        // a __MACOSX gets deleted because its content was deleted
+        // it will will simply not be a directory any more and not cause trouble
```
rvanheest:
```suggestion
        // it will simply not be a directory any more and not cause trouble
```

PR #176, opened 2019-08-23
EASY-2235: Include INITIAL_HEAP_SIZE java cmd argument

lindareijnhoudt:
Fixes EASY-2235

#### When applied it will
* add a new .conf file to set the `-Xmx` argument with reasonable default `64m`

easy-dtap                      | [PR#374](https://github.com/DANS-KNAW/easy-dtap/pull/374)     | ..


src/main/rpm/2-post-install.sh
```diff
@@ -27,7 +27,7 @@ PHASE="POST-INSTALL"
 
 echo "$PHASE: START (Number of current installations: $NUMBER_OF_INSTALLATIONS)"
 service_install_initd_service_script "$INSTALL_DIR/install/$MODULE_NAME-initd.sh" $MODULE_NAME
-service_install_systemd_unit "$INSTALL_DIR/install/$MODULE_NAME.service" $MODULE_NAME "$INSTALL_DIR/install/override.conf"
+service_install_systemd_unit "$INSTALL_DIR/install/$MODULE_NAME.service" $MODULE_NAME "$INSTALL_DIR/install/charset.conf"
```
lindareijnhoudt: @janvanmansum should the memusage.conf also be listed here?

src/main/assembly/dist/install/memusage.conf
```diff
@@ -0,0 +1,2 @@
+[Service]
+Environment="MAX_HEAP_SIZE=-Xmx64m"
```
rvanheest: newline at the end avoids unnecessary diffs in the future

rvanheest (commented):
I'm not familiar enough with systemd to give the green light here. As far as I understand it, you're adding a new configuration for 'memory usage'? My main question is: "**why can't you add that to the existing configuration?**"

lindareijnhoudt:
> I'm not familiar enough with systemd to give the green light here. As far as I understand it, you're adding a new configuration for 'memory usage'? My main question is: "**why can't you add that to the existing configuration?**"

These conf-files can contain multiple `Environment` vars, but changing it in dtap, it will replace the entire `Environment` var, so it will loose what was already there.
Therefor i added a new conf-file, with only 1 Environment var.
Installing the rpm will use all conf-files from this directory

PR #188, opened 2019-12-06
EASY-2400 update documentation

rkruithof:
Fixes EASY-2400

#### When applied it will
* update documentation to be in line with other projects
* remove references to outdated files from bin.xml

### DIFF
--- a/docs/api/api.yml
+++ b/docs/api/api.yml
@@ -1,13 +1,11 @@
 openapi: 3.0.1
 
 servers:
-  - description: Debug environment
     url: http://localhost:20190/
-  - description: Local Vagrant machine
-    url: http://test.dans.knaw.nl
+
 info:
   description: Public interface for depositing datasets in EASY.
-  version: 1.0.0
+  version: 5.1.2
   title: Deposit API
   contact:
     email: info@dans.knaw.nl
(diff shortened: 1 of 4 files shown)


docs/api/api.yml
```diff
@@ -1,13 +1,11 @@
...
 servers:
-  - description: Debug environment
     url: http://localhost:20190/
-  - description: Local Vagrant machine
-    url: http://test.dans.knaw.nl
+
 info:
   description: Public interface for depositing datasets in EASY.
-  version: 1.0.0
+  version: 5.1.2
```
rvanheest: what does this version mean?

rkruithof: I thought this version number was supposed to be the version number of the deposit-api so you can see if the api spec you are looking at is up to date with the software. 
Updating it manually like this may result in it falling behind though.

janvanmansum: Good point. I hadn't really considered it yet. Couple of points:
* 5.1.2 is the version of the parent POM, so in any case this is the wrong version number
* The [OpenAPI specs say](https://swagger.io/specification/#infoObject) about this field: "The version of the OpenAPI document (which is distinct from the OpenAPI Specification version or the API implementation version)". In other words, they are assuming that we are versioning our API. The API version would not automatically be the same as the modules version. The module can go through several versions without the API changing. I am not sure yet what the best policy is here. In this particular case I would change it back to `1.0.0`, as we haven't released this module yet, and the API will initially be version 1.0.0 in any case. @rvanheest @lindareijnhoudt : we should probably think about a general policy with respect to how to fill in this field.

PR #205, opened 2020-01-24
EASY-2502: restoring/refactoring tests - phase 2 SubmitJob + fixed a leak

jo-pol:
Fixes EASY-2502: restoring/refactoring tests - phase 2 SubmitJob + fixed a leak

#### When applied it will
* fix leaking sensitive information (in the status message for the client) when the background thread fails to complete a `SubmitJob`. The message now will have a "please contact DANS" message in case of problems
* have tests for `SubmitJob` in `SubmitterSpec`, using the `Submitter` was the easiest way to prepare test for `SubmitJob`.
* 


src/main/scala/nl.knaw.dans.easy.deposit/StateManager.scala
```diff
@@ -84,8 +85,10 @@ case class StateManager(draftDeposit: DepositDir, submitBase: File, easyHome: UR
       }
   }
 
-  def setStateFailed(description: String): Try[Unit] = Try {
-    saveNewStateInDraftDeposit(StateInfo(State.submitted, description))
+  def setMailToDansDescription(state: State): Try[Unit] = Try {
+    // using saveNewStateInDraftDeposit could inadvertently change the state
+    draftProps.setProperty(stateDescriptionKey, mailToDansMessage)
+    draftProps.save()
```
rvanheest: Why this `state: State` parameter? It's currently not used. From the usage I would expect a `draftsProps.setProperty(..., state)` somewhere here, which makes it equivalent to the original implementation.

src/main/scala/nl.knaw.dans.easy.deposit/SubmitJob.scala
```diff
@@ -59,12 +62,16 @@ class SubmitJob(draftDepositId: UUID,
...
-          .doIfFailure { case e => logger.error(s"[$draftDepositId] could not set state to FAILED after submission failed", e) }
+        draftDepositStateManager.setMailToDansDescription(State.submitted)
+          .doIfFailure { case e => logger.error(s"[$draftDepositId] could not set state description after submission failed", e) }
       case Success(()) =>
         sendEmail
           .doIfSuccess(_ => logger.info(s"[$draftDepositId] finished with dispatched submit action"))
-          .doIfFailure { case e => logger.error(s"[$draftDepositId] deposit submitted but could not send confirmation message", e) }
+          .doIfFailure { case e =>
+            // applying setMailToDansDescription (with something like no confirmation message sent)
+            // could get overwritten by easy-ingest-flow, or worse: overwrite a change by easy-ingest-flow.
```
rvanheest: This is not true.

1. `draftDepositStateManager.setMailToDansDescription` is not supposed to be called here. The fact that sending an e-mail failed is no reason for the state to change.
2. the draft deposit is NEVER overwritten by `easy-ingest-flow` OR ANY OTHER APPLICATION. The draft deposits are managed by `easy-deposit-api` only. `easy-ingest-flow` has its own inbox. `easy-manage-deposit` updates the state of the deposit in `easy-ingest-flow` on publishing/rejecting a dataset, but it will NEVER touch the draft deposits in `easy-deposit-api`. The state of a draft deposit is only changed by `easy-deposit-api` (see `StateManager`).

I would suggest to remove this comment and restore the original code.

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -76,22 +76,20 @@ class Submitter(submitToBaseDir: File,
                  else ""
       _ = logger.info(s"[$depositId] dispatching submit action to threadpool executor$queueMsg")
       _ <- jobQueue.scheduleJob {
-        new SubmitJob(
-          draftDepositId = draftDeposit.id,
-          depositUiURL = depositUiURL,
-          groupPrincipal = groupPrincipal,
+        new SubmitJob(draftDepositId = draftDeposit.id,
```
rvanheest: add the newline back in, makes the code more readable.

src/main/scala/nl.knaw.dans.easy.deposit/Submitter.scala
```diff
@@ -76,22 +76,20 @@ class Submitter(submitToBaseDir: File,
...
           msg4DataManager = datasetMetadata.messageForDataManager.getOrElse("").stripLineEnd.toOption,
           agreementData = AgreementData(user, datasetMetadata),
           draftDepositStateManager = draftDepositStateManager,
+          groupPrincipal = groupPrincipal,
+          depositUiURL = depositUiURL,
+          submitDir = submitDir,
           agreementGenerator = agreementGenerator,
-          mailer = mailer,
-        )
+          mailer = mailer)
```
rvanheest: Add the comma and the newline before the `)` back in. Makes future diffs smaller and less confusing.

src/main/scala/nl.knaw.dans.easy.deposit/SubmitJob.scala
jo-pol: `set...Description` does not change the state and it should not. The method that does change the state checks for a valid transition. 

It just changes the description when the background thread fails. No email is more or less a failure in the background thread, but in this case ingest-flow will start processing. "overwritten by" is a shortcut for "when the api peeks into ingest-flow to update the state".

I somehow was thinking whether we could tell the end user somehow that the confirmation mail was not sent.

rvanheest:
Results of review with @jo-pol in https://github.com/DANS-KNAW-jp/easy-deposit-api/pull/10

PR #209, opened 2020-02-06
Refactor ddm

jo-pol:
Fixes EASY-

#### When applied it will (_SQUASH MERGE MESSAGE !_)
* no longer validate the scheme of author IDs
* use defaults for date/relation qualifiers
* use zero as default when some coordinate parts are missing, such as `x` without `y` or vice versa
* still reject unknown fields and duplicate date-created, date-available
* still validate Enum values: RelationQualifier, DateQualifier, State, AccessRights, PrivacySensitiveDataPresent
* renamed methods: 
  * `withNonEmpty` -> `withValue`
  * `nonBlankOrNull` -> `orOmit` (for XML attribute)
  * `nonBlankOrEmpty` -> `orEmpty` (for text value of XML element)
  * `details` -> nestedXML / basicContent (pattern matches in the latter)
* have logic of DDM object moved to lazy val-s (no more def's) of objects in `dm` package (value objects for json structure) this reduced the number/complexity of cases in pattern matches
* additional/refactored unit tests
* 

* [x] mapping in `DDM`


rvanheest:
build failure in Travis due to a missing license header

src/main/scala/nl.knaw.dans.easy.deposit/docs/CollectionUtils.scala
```diff
@@ -31,7 +32,16 @@ object CollectionUtils {
   }
 
   implicit class RichOptionSeq[T](val sources: Option[Seq[T]]) extends AnyVal {
+
+    /** filters elements without a value (before mapping) */
     def withNonEmpty: FilterMonadic[T, Seq[T]] = sources.toSeq.flatten.withNonEmpty
+
+    /** filters empty XML elements after mapping (in case another field is mapped to a value) */
+    def toNonEmptyMap(f: T => Elem): Seq[Elem] = {
```
rvanheest: 1. There is only one use case for this. Are you planning on using this more often? Otherwise abstracting it into here may be a bit early.
2. I'm not happy with the name of this function `toNonEmptyMap` assumes you're returning a `Map` that has a type-level guarantee that it is not empty.
3. Why can't you use the usual `withNonEmpty.map(...)` instead in this single use case?

src/main/scala/nl.knaw.dans.easy.deposit/docs/DDM.scala
```diff
@@ -100,62 +91,55 @@ object DDM extends SchemedXml with DebugEnhancedLogging {
...
+        <label scheme={ rel.schemeOrNull } href={ rel.urlOrNull }>{ rel.valueOrThrow }</label>
+      case rel: Relation => // Relation(_,None,None) is skipped so we do have a title (via value) and therefore a language
+        <label xml:lang={ lang } href={ rel.urlOrNull }>{ rel.valueOrThrow }</label>
+    }).withLabel(qualifier(cleanRelation))
+  }
 
-  private def details(source: SchemedKeyValue, label: String, lang: String): Elem = {
-    (label, source) match {
+  private def qualifier(cleanRelation: RelationType) = {
+    if (cleanRelation.url.isEmpty) cleanRelation.qualifierOrThrow
```
rvanheest: this `isEmpty` is confusing me. It might be from `Option.isEmpty` or `String.isEmpty`.
how about:

```scala
cleanRelation.url
  .map(_ => cleanRelation.qualifierOrThrow.replace("dcterms", "ddm"))
  .getOrElse(cleanRelation.qualifierOrThrow)
```

src/main/scala/nl.knaw.dans.easy.deposit/docs/dm/Spatial.scala
```diff
@@ -25,35 +25,43 @@ object Spatial {
...
       case Some("RD") => Spatial.RD_SRS_NAME
       case Some(s) if s.trim.nonEmpty => s
-      case _ => null // will suppress the XML attribute
+      case _ => throw new IllegalArgumentException(s"scheme is mandatory for $errorDetails")
     }
   }
+
+  private def errorDetails = s"${ getClass.getSimpleName }: ${ toJson(this) }"
+
+  protected def allOrNone(msg: String, args: Option[String]*): Unit =
```
rvanheest: please annotate this and all use cases with an `@throws` if you don't want to return a `Try`

src/test/scala/nl.knaw.dans.easy.deposit/docs/DDMSpec.scala
```diff
@@ -180,52 +183,116 @@ class DDMSpec extends TestSupportFixture with DdmBehavior {
...
+
+  "multiple datesCreated" should "fail" in {
+    the[IllegalArgumentException] thrownBy new DatasetMetadata(
+      dates = Some(mandatoryDates ++ mandatoryDates),
+    ) should have message
+      """requirement failed: At most one allowed; got [{"scheme":"dcterms:W3CDTF","value":"2018-05-29","qualifier":"dcterms:created"},{"scheme":"dcterms:W3CDTF","value":"2018-05-29","qualifier":"dcterms:created"}]"""
+  }
+
+  "metadata without datesCreated" should "fail" in {
+    DDM(new MinimalDatasetMetadata( // TODO DDM would not validate
```
jo-pol: This test fails. As the UI is supposed to provided all mandatory values, this is not a direct production problem.

PR #216, opened 2020-03-19
EASY-1945: upgrade dependencies

pfrolke:
Fixes EASY-1945

#### When applied it will
* upgrade parent POM version
* remove self-contained dependency version tags
* move unique dependency versions that aren't included in a parent POM to properties
* fix bugs caused by dependency breaking changes

#### TODO:
- [x] fix incorrect conversion of `private val` params in `DatasetMetadata.scala` caused by `json4s` version > 3.6.

#### related pull requests on github
repo                       | PR
-------------------------- | -----------------
easy-                      | [PR#](PRlink)


src/main/scala/nl.knaw.dans.easy.deposit/authentication/TokenSupport.scala
```diff
@@ -56,7 +59,7 @@ trait TokenSupport extends DebugEnhancedLogging {
   def decodeJWT(token: String): Try[AuthUser] = {
     for {
       decoded <- Jwt.decode(token, tokenConfig.secretKey, Seq(tokenConfig.algorithm))
-      parsed <- fromJson(decoded)
+      parsed <- fromJson(decoded.toJson)
```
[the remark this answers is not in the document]
jo-pol: in the toJson I see

    "iss" -> issuer,
    "sub" -> subject,
    "aud" -> audience.map(set => if (set.size == 1) set.head else set),
    "exp" -> expiration,
    "nbf" -> notBefore,
    "iat" -> issuedAt,
    "jti" -> jwtId

Which should parse as `case class Token(exp: Long, iat: Long, uid: String)` 
I can match exp/iat but not uid, otherwise the the detour could be rewritten as
`Jwt.decode(...).map(claim => Token(claim.exp, claim.iat, claim.???))`

pom.xml
```diff
@@ -32,6 +32,8 @@
     <inceptionYear>2018</inceptionYear>
     <properties>
         <main-class>nl.knaw.dans.easy.deposit.Command</main-class>
+        <jwt-core.version>4.3.0</jwt-core.version>
```
jo-pol: I do see commits with 4.3.0 but the project readme and last [release](https://github.com/pauldijou/jwt-scala/releases) are still at 4.2.0. Don't know what the implies.

pfrolke: Hmm, mvnrepository.com does list 4.3.0 as the latest version https://mvnrepository.com/artifact/com.pauldijou/jwt-core

src/main/scala/nl.knaw.dans.easy.deposit/authentication/TokenSupport.scala
jo-pol:
```suggestion
      parsed <- fromJson(decoded.content)
```
change also 
https://github.com/DANS-KNAW/easy-deposit-api/blob/1713f181141efdbb526554175970f1ad871f83a2/src/main/scala/nl.knaw.dans.easy.deposit/authentication/TokenSupport.scala#L74
into
```
That should fix `SessionSpec` 
private case class Token(uid: String)
```

PR #218, opened 2020-04-02
EASY-2179: update scalatra

pfrolke:
fixes EASY-2179

#### When applied it will
* update parent POM version
* fix deprecation warnings
* fix scalatra breaking changes

#### related pull requests on github
repo                       | PR                | Note
easy-                      | [PR#](PRlink)     |

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>nl.knaw.dans.shared</groupId>
         <artifactId>dans-scala-app-project</artifactId>
-        <version>6.0.2-SNAPSHOT</version>
+        <version>6.0.2</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <groupId>nl.knaw.dans.easy</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>nl.knaw.dans.shared</groupId>
         <artifactId>dans-scala-app-project</artifactId>
-        <version>6.0.1</version>
+        <version>6.0.2-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <groupId>nl.knaw.dans.easy</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
     <parent>
         <groupId>nl.knaw.dans.shared</groupId>
         <artifactId>dans-scala-app-project</artifactId>
-        <version>6.0.2</version> 
+        <version>6.0.2</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
     <groupId>nl.knaw.dans.easy</groupId>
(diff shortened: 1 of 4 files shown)


src/main/scala/nl.knaw.dans.easy.deposit/servlets/package.scala
```diff
@@ -194,7 +194,7 @@ package object servlets extends DebugEnhancedLogging {
     private def moveIfNotAnArchive(srcItem: FileItem, multiPartLocation: File, targetDir: File, id: UUID): Try[Unit] = {
       val target = targetDir / srcItem.name
       logger.info(s"[$id] staging upload to $target, contentType = ${ srcItem.contentType }")
-      if (srcItem.name.isBlank) Success(()) // skip form field without selected files
+      if (srcItem.name.isEmpty) Success(()) // skip form field without selected files
```
jo-pol, line 197: Please examine the difference between both methods: `isEmpty` checks for an empty array, `isBlank` checks for an empty string after trimming white space. 
So revert and replace `import  org.scalatra.util.RicherString._` with `import nl.knaw.dans.lib.string._`

However @rvanheest doesn't scalatra's motivation to deprecate also apply to our own library? 

    @deprecated("RicherString#isBlank has been deprecated, it conflicts with Java 11's String#isBlank", "2.7.0")

Might apply to a higher Java version than we are using.

src/main/scala/nl.knaw.dans.easy.deposit/servlets/package.scala
```diff
@@ -222,7 +222,7 @@ package object servlets extends DebugEnhancedLogging {
 
     private def skipLeadingEmptyFormFields(): Unit = {
       // forward pointer after check
-      while (fileItems.headOption.exists(_.name.isBlank)) { fileItems.next() }
+      while (fileItems.headOption.exists(_.name.isEmpty)) { fileItems.next() }
```
jo-pol, line 225: Same here

pom.xml
```diff
@@ -20,7 +20,7 @@
     <parent>
         <groupId>nl.knaw.dans.shared</groupId>
         <artifactId>dans-scala-app-project</artifactId>
-        <version>6.0.1</version>
+        <version>6.0.2</version>
```
rvanheest, line 23:
```suggestion
        <version>6.0.2</version>
```

src/main/scala/nl.knaw.dans.easy.deposit/servlets/package.scala
rvanheest: @jo-pol since we're still running on Java 8 this isn't really a concern for us right now. It would be once we migrate to a higher version of Java. This is quite unlikely for now, since Scala 2.13 (the newer version that we haven't yet gotten around to upgrading to) still supports Java 8. So let's not bother with this for now and keep the `.isBlank` from our own library.

```suggestion
      if (srcItem.name.isBlank) Success(()) // skip form field without selected files
```

src/main/scala/nl.knaw.dans.easy.deposit/servlets/package.scala
rvanheest: 👍 
```suggestion
      while (fileItems.headOption.exists(_.name.isBlank)) { fileItems.next() }
```

